PR #2167, opened 2026-04-09
Stop leaking Lightpanda CDP sessions on worker restarts

viktor-shcherb:
## Summary

Three layered fixes for the session-leak storm that drained the Lightpanda monthly browser-hours quota in ~22 minutes today, breaking Starbucks and every other CDP-routed board.

Evidence from the Lightpanda dashboard (13 sessions opened in 22 min, **12 of 13 ran the full ~6 min idle timeout with 0 bytes transferred** — a smoking-gun for "process exited without closing the session"):

```
21:18:56 / 21:18:59  pair, both 6m10s, 0 B
21:25:08 / 21:25:11  pair, both 6m10s, 0 B
21:31:22 / 21:31:26  pair, both 6m07s, 0 B
21:37:35 / 21:37:38  pair, both 6m07s, 0 B
21:27:47             1.023s — the only one closed cleanly
... (8 more 6-min orphans)
```

### 1. ``shutdown_all_sessions()`` was never wired in (`cli.py`)

The function existed in `shared/cdp.py` but no shutdown path called it. `LightpandaTransport.aclose()` is intentionally a no-op (decoupled from httpx client lifecycle), so even on a clean SIGTERM the playwright/CDP websocket stayed open and Lightpanda held the orphaned session for ~6 min idle timeout.

**Fix:** call `shutdown_all_sessions()` in the `cli.py run()` `finally:` block, before `close_all_pools()`. Every clean exit now releases the session immediately.

### 2. Reset-on-every-error tore the session down on benign failures (`shared/cdp.py`)

`_LightpandaSession.request()` had a blanket `except Exception → _reset_locked()`. A single 404 from the target site (or any per-request error) tore down the entire CDP session. The next call would reconnect — and any process exit during that reconnect leaked another orphaned 6-minute session.

**Fix:** introduce `_is_connection_error(exc)` that matches conservative substrings against `str(exc)` (`"target closed"`, `"websocket"`, `"browser closed"`, `"connect_over_cdp"`, `"disconnected"`, …). The session is reset **only** on those — per-request timeouts and 4xx/5xx leave the session intact.

### 3. Redis pool too small → worker crash loop → restart leaks (`config.py`)

`redis_max_connections` defaulted to **20**, but production runs `DISCOVERY_CONCURRENCY=30 + MONITOR_CONCURRENCY=10 = 40` concurrent worker tasks per process, all calling `claim_work()` against the same shared pool. The 21st concurrent claim crashed with `redis.exceptions.MaxConnectionsError: Too many connections`.

The pipeline catches the exception (`pipeline.claim_error` warning + back-off), so the *worker task* survives — but the underlying churn caused **28 worker container restarts in one day** on `deploy-worker-1-1` today, each restart leaking a Lightpanda session via bug #1.

**Fix:** bump default to **60** (40 needed + headroom for ad-hoc Redis ops + reschedule bursts). Single-line config change; no env var update needed since the new default is tighter than the production env requires.

## Test plan

- [x] `uv run pytest tests/test_cdp_transport.py` → **65 passed** (19 new)
- [x] `uv run pytest tests/` → **3068 passed, 37 skipped** (no regressions)
- [ ] Post-merge: redeploy crawler, monitor for `cdp.session.opened` event count drop and absence of `pipeline.claim_error / MaxConnectionsError`
- [ ] Post-merge: confirm next worker restart cleanly releases Lightpanda sessions (browser-hours dashboard)

## New tests

- `TestIsConnectionError` — 13 cases covering connection-error markers (real playwright messages) vs benign errors (timeouts, 404s, decoding errors, ECONNRESET — which is *not* on the marker list and won't reset)
- `TestSessionResetNarrowing` — 3 cases proving benign errors don't reset the session, real connection errors do, and a 404 response (which playwright returns as a Response, not an exception) leaves the session intact
- `TestShutdownAllSessions` — 3 cases for the wired shutdown helper: closes every cached session, swallows close errors so a bad session doesn't block the others, and is a no-op on empty cache
- `TestRedisPoolSizingDefault` — regression guard on the pool default vs production worker concurrency


viktor-shcherb:
## Critic-pass follow-up (commit cedc7df3)

Spawned three independent critic agents (correctness, completeness, observability) on the previous commit. They surfaced four real gaps that would let the bug recur. All four are now patched in the new commit, each with a regression test.

### What the critics found

**[CRITICAL] Workspace ws commands leaked sessions on every invocation** — `src/workspace/commands/crawl.py` calls `create_http_client()` in 7 `_run()` paths and only awaited `http.aclose()` in their finally blocks. Each ws command is a one-shot process — no cli.py finally fires — so every `ws probe` against a CDP-routed host opened a Lightpanda session that lived until the ~6-minute idle reap. **Fix:** new `_shutdown_http(http)` helper that closes the http client AND calls `shutdown_all_sessions()`; routed all 7 sites through it. Regression test: `TestWorkspaceShutdownWiring::test_ws_crawl_uses_shutdown_helper` asserts only one direct `http.aclose()` exists in `crawl.py` (the one inside the helper).

**[HIGH] `shutdown_all_sessions` could hang the cli finally forever** — sequential awaits with no per-session timeout. A real Lightpanda websocket close can hang on network glitches; if cli.py finally then blocks past docker stop grace period, container gets SIGKILL'd and ALL sessions leak. **Fix:** per-session `asyncio.wait_for(timeout=5s)` + `asyncio.gather` for parallel close. Total shutdown bounded to ~5s regardless of session count. Regression test: `test_hung_close_does_not_block_others` proves hung session times out and good ones close in <1s in parallel.

**[HIGH] No `stop_grace_period` in docker-compose** — defaults to 10s. With Postgres pool drain + Lightpanda close + workers finishing in-flight tasks under load, 10s can be too tight. SIGKILL after grace period leaks sessions. **Fix:** `stop_grace_period: 30s` on every crawler service (worker-1/2/3, browser-1, exporter, drain).

**[HIGH] No regression test asserted cli.py actually called the helper** — a future refactor that split cli.py, removed the import, or moved the call out of the finally would silently reintroduce today's bug, and CI would pass. **Fix:** `TestCliShutdownWiring` (3 tests) + `TestWorkspaceShutdownWiring` (1 test) — source-level guards on `cli.run()` and `crawl.py`.

### What was deferred to a follow-up PR

The observability critic identified valuable additions that don't fix the bug but make regressions detectable in <5 min instead of after quota exhaustion:

- `cdp_sessions_active` Gauge metric (incremented in `_ensure_open`, decremented in `_reset_locked`/`close`)
- `cdp_request_errors_total` Counter labeled by `category={connection,per_request}`
- Grafana panel on `cdp_sessions_active` and `rate(cdp_request_errors_total[5m])`
- Alert on `increase(cdp_sessions_active[15m]) > 2` (sessions accumulating without closing)

These touch `metrics.py` + `grafana-dashboard.json` + alerting config — separate concern, separate PR.

### What I rejected as not-a-bug

- **`asyncio.CancelledError` not caught in `request()`** (correctness critic) — In practice, `pipeline.py`'s worker loop polls `shutdown_event.is_set()` between tasks rather than cancelling in-flight ones, so CancelledError doesn't actually fire during normal shutdown. Adding `except BaseException` wouldn't change observed behavior, and it'd risk masking other intentional cancellations elsewhere.
- **Substring marker false positives** (e.g., `"websocket"` matching `"WebSocket established"`) — false positives just trigger an unnecessary session reopen, which is a strict improvement over the pre-fix "reset on every error". Tightening the markers is a low-priority polish that can wait.

### Test results

- 70 cdp tests pass (was 65; +5 new)
- 3073 / 3110 full crawler suite (37 skipped, no regressions)

viktor-shcherb:
## Second critic pass (commit 0ae28f96)

Spawned three more critic agents (correctness, race-conditions, leak-path-coverage) on the previous commit. Findings + actions:

### Confirmed safe
- **Leak-path coverage critic: ZERO uncovered leak paths in production.** It enumerated all `create_http_client()` callers, all `LightpandaTransport(` instantiations, all `_get_session(` callers, all `build_cdp_mounts(` callers, the `crawler board` debug command path, scripts, conftest, background tasks, and other cli subcommands. Defense-in-depth confirmed: every CDP-routed code path has both an inner http close AND an outer `shutdown_all_sessions()` flush (either via cli.py finally or via `_shutdown_http` helper).

### "Race condition" was a false alarm
Two critics flagged the snapshot/gather/clear pattern in `shutdown_all_sessions` as racing with concurrent session creation. **It's unreachable in production:** `run_pipeline` uses `asyncio.TaskGroup` which awaits all worker tasks before returning, and `_discovery_worker` only exits its loop AFTER finishing its current task. By the time cli.py's outer finally fires, no worker can still be creating sessions. Documented as a non-bug, no fix needed.

### Real findings patched in this commit

1. **`_shutdown_http` could mask the http.aclose exception.** If both `http.aclose()` and `shutdown_all_sessions()` raise, Python's try/finally semantics suppress the original. Fix: wrap the inner `shutdown_all_sessions()` call in its own try/except + log instead of propagating, so the more useful diagnostic always wins.

2. **No autouse conftest fixture isolating `cdp._sessions` / `cdp._SHUTDOWN_CLOSE_TIMEOUT_SEC` between tests.** Tests currently rely on per-test `finally:` cleanup. A future test that forgets it would silently leak module state — exactly the kind of bug that only shows up in CI on specific test orderings. Fix: add an autouse function-scoped fixture in `tests/conftest.py` that resets both. Lazy import so non-CDP tests don't pull playwright/httpx.

3. **`test_ws_crawl_uses_shutdown_helper` was fragile to renames.** The previous version asserted `source.count("await http.aclose()") == 1`. A future PR that renamed the local variable from `http` to `client` would break the test for the wrong reason. Fix: parse line-by-line, skip the helper definition, assert no `.aclose()` call (any name) lives outside the helper. Plus a separate `call_sites >= 5` check to catch the inverse: helper exists but isn't being used.

### Findings NOT patched (with rationale)

- **`asyncio.wait_for` cancellation semantics on Playwright close** — theoretical, no evidence of real problem in practice. The 5s timeout is a hedge against pathological cases.
- **`_ensure_open()` lock contention not exercised by tests** — would need a real CDP fake server, significant scope creep for an integration test that wouldn't catch a real bug pattern.
- **Healthcheck firing during stop_grace_period** — Docker doesn't restart on healthcheck failures during shutdown, so it's just log noise. Low impact.
- **Test timing flake risk on `test_hung_close_does_not_block_others`** — 10× headroom (asserts <1s with ~0.1s expected) is generous; not a realistic flake source.

### Tests
- 70 cdp tests pass (no new tests this commit; the existing `TestWorkspaceShutdownWiring` exercises the hardened source-parsing logic)
- 3073 / 3110 full crawler suite

Issue #2214, opened 2026-04-17
Workday/Deloitte/Uber scrape bot-blocks (403/406) — 5.4k errors/12h


viktor-shcherb:
## Problem

Per-posting scrape fetches are being bot-blocked at the JSON-endpoint layer. Over the past 12h this accounts for **5,442 errors (~83% of all crawler errors)**, nearly all on Workday-backed career sites plus Deloitte and Uber.

## Evidence (Loki, past 12h, 2026-04-17)

**HTTP 403 (4,622 total)** — all `batch.scrape.error`:

| host | count |
|------|-----:|
| `wf.wd1.myworkdayjobs.com` (Wells Fargo) | 966 |
| `citi.wd5.myworkdayjobs.com` | 372 |
| `hpe.wd5.myworkdayjobs.com` | 296 |
| `globalhr.wd5.myworkdayjobs.com` | 263 |
| `jj.wd5.myworkdayjobs.com` (J&J) | 248 |
| `medtronic.wd1.myworkdayjobs.com` | 216 |
| `ag.wd3.myworkdayjobs.com` | 189 |
| `ms.wd5.myworkdayjobs.com` | 169 |
| `abb.wd3.myworkdayjobs.com` | 141 |
| `apply.deloitte.com` | 141 |
| `target.wd5.myworkdayjobs.com` | 104 |
| `gdit.wd5.myworkdayjobs.com` | 103 |
| `careers.loreal.com` | 94 |
| `amat.wd1.myworkdayjobs.com` | 94 |
| `mastercard.wd1.myworkdayjobs.com` | 82 |
| … 30+ other Workday hosts | ~1,050 |

Sample error:

```
Client error '403 Forbidden' for url
  'https://wf.wd1.myworkdayjobs.com/wday/cxs/wf/WellsFargoJobs/job/.../Senior-Product-Manager_R-521988-1'
```

Deloitte 403s redirect to `/en_US/careers/Error` — classic anti-bot landing.

**HTTP 406 (820 total)**:

| host | count |
|------|-----:|
| `www.uber.com` | 809 |
| `apply.deloitte.com` | 11 |

Sample Uber 406:
```
Client error '406 Not Acceptable' for url
  'https://www.uber.com/careers/list/155842'
```

406 on a plain GET typically means the server rejects the `Accept` header or is fingerprinting the request.

## Hypotheses

[...]

PR #2232, opened 2026-04-19
Force Accept: application/json in Lever monitor (fixes #2228)

### LINKED ISSUE #2228: Lever monitor returning HTML instead of JSON — 123 errors/h on jobs.lever.co
## Problem

The Lever monitor is failing for ~15 boards on `jobs.lever.co` with a JSON parse error — the response body is HTML (likely a block/maintenance page), but we try to decode it as JSON and the caller just sees a confusing `Expecting value: line 1 column 1 (char 0)`.

## Evidence

Current rate: **123 `batch.monitor.error`/h on `jobs.lever.co`** (+ 4 on `jobs.eu.lever.co`). Dominant contributor to the ~3% monitor-failure rate today (04-18 mean is up to 3.0% vs. 1.2–1.6% earlier in the week).

Sample:

```json
{"board_url":"https://jobs.lever.co/veeva",
 "crawler_type":"lever",
 "error":"Expecting value: line 1 column 1 (char 0)",
 "event":"batch.monitor.error",
 "exception":"Traceback ... src/core/monitor.py:334 in monitor_one_stream\n  async for batch in stream_fn(board, http, pw=pw):"}
```

Affected tenants observed: `veeva`, `unify`, `sonarsource` (and more — full list via Loki label `board_id`).

## Hypotheses

1. Lever is returning an HTML block/captcha page to certain requests (UA fingerprinting, IP-based, or rate-limit).
2. These specific tenants switched their public Lever feed off or are on Lever's new careers surface which doesn't serve the same JSON shape.

## Suggested fix

- In the Lever monitor, detect non-JSON responses (content-type sniff or leading `<`) and emit a **structured** error (`workable.rate_limited`-style) with the actual snippet, instead of letting `json.loads` raise.
- Once we can see the real body, classify: block page vs. tenant removed vs. Lever API change.
- If it's a block, share the backoff/rotation story with the Workable `rate_limited` path.

## Acceptance

- `batch.monitor.error` on `jobs.lever.co` drops ≥90%.
- When Lever blocks, logs show the response body/status, not a bare JSON decode error.

viktor-shcherb:
## Summary

Adds `Accept: application/json` to the three Lever API call sites. Fixes the regression where `api.lever.co` started returning a minimal HTML widget instead of JSON after commit c204b210 flipped the default client Accept header to a browser-style value.

## Root cause

`api.lever.co/v0/postings/{token}` is content-negotiated. Reproducible in 3 seconds:

```bash
# Returns 10,946 bytes of JSON
curl -sS -H 'Accept: application/json' 'https://api.lever.co/v0/postings/ledger?limit=100' | head -c 100

# Returns 250 bytes of HTML — the "Derby App" widget
curl -sS -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' \
    'https://api.lever.co/v0/postings/ledger?limit=100'
```

The browser Accept header was added in c204b210 on 2026-04-17 to unblock Uber (#2214). Lever's Derby widget is intentional — it serves an HTML embed at the same path for customers who want a drop-in careers list. The crawler always needs JSON.

When the HTML comes back, `response.raise_for_status()` passes (200 OK) and `response.json()` raises a bare `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` — no status, no content-type, no body snippet — from `src/core/monitors/lever.py:178`.

## Timeline (Loki)

Lever `JSONDecodeError` events aligned exactly with the c204b210 deploy:

| date | count |
|---|---:|
| 2026-04-15 | 1 |
| 2026-04-17 | 627 |
| 2026-04-18 | 2,847 |
| 2026-04-19 (partial) | 1,254 |

## Fix

Single-header override on the three call sites in `lever.py` (`discover`, `_probe_token`, `_fetch_job_count`). httpx merges per-request headers over client defaults, so `Accept: application/json` wins.

No need to change `shared/http.py` or other monitors — I checked the other JSON-API monitors (`greenhouse`, `ashby`, `recruitee` per-tenant, `gem`, `pinpoint`, `dvinci`) against the same header test: none of them flip response shape based on Accept. Lever is the only known case.

## Verification

Live smoke test against three affected boards via the default crawler HTTP client (same browser Accept header used in production):

```
ledger: 1 jobs (first: 'Local Market Outreach - Australia')
veeva: 1063 jobs (first: '(Associate) Team Lead - Data Operations')
activecampaign: 21 jobs (first: 'Account Executive 1')
```

All three match the counts observed via `curl -H 'Accept: application/json'`.

## Test plan

- [x] `pytest -k lever` — 60 passed
- [x] Live `discover()` against ledger, veeva, activecampaign returns correct job counts
- [ ] After deploy: `batch.monitor.error{crawler_type="lever"}` drops to near zero in Grafana

## Alternatives considered

1. **Emit structured error instead of bare JSONDecodeError** — worth doing eventually so we can diagnose the next Lever-shape-shift faster, but a separate PR. This fix is the minimum to stop the bleed.
2. **Revert the default Accept header** — would re-break Uber. The browser Accept is correct for HTML scrapers; API monitors should opt into `application/json` explicitly.
3. **Flip the default Accept to `*/*` for API-like paths** — too fragile; routing by URL shape inside the client layer is harder to reason about than one header on three lines.


viktor-shcherb:
Relentlessly reverified against the top 25 boards still emitting `JSONDecodeError` in Loki over the last 12 hours.

### Direct Accept-header A/B

Issued the same GET to `api.lever.co/v0/postings/{tok}?limit=100` with each Accept header:

| | old browser Accept | new `application/json` |
|---|---:|---:|
| JSON response | **0 / 25** | **24 / 25** |
| HTML (Derby widget) | **23 / 25** | 0 / 25 |
| network error | 2 / 25 | 1 / 25 (transient) |

Representative per-token results:

```
agicap                 HTML(<!DOCTYPE html>)  → JSON(42)
aleph                  HTML(<!DOCTYPE html>)  → JSON(91)
anchorage              HTML(<!DOCTYPE html>)  → JSON(94)
binance                HTML(<!DOCTYPE html>)  → JSON(100)
pipedrive              HTML(<!DOCTYPE html>)  → JSON(46)
sonarsource            HTML(<!DOCTYPE html>)  → JSON(100)
...
```

### Full pipeline (patched monitor + production HTTP client)

Ran `lever.discover()` through `src/shared/http.create_http_client()` — same defaults as the live worker:

```
✓ agicap                    42 jobs
✓ aleph                     91 jobs
✓ anchorage                 94 jobs
✓ anybotics                 23 jobs
✓ anyscale                  1 jobs
✓ arbitalhealth             2 jobs
✓ atomcomputing             21 jobs
✓ binance                   365 jobs
✓ crypto                    43 jobs
✓ daedalean                 8 jobs

Full-pipe: 10 OK, 0 failed
```

Every board is now returning a JSON list, with counts consistent with direct `curl -H 'Accept: application/json'` against the same endpoint.

PR #2234, opened 2026-04-19
Expand robots.txt disallow to locale-prefixed paths

viktor-shcherb:
## Summary
- `robots.txt` disallow entries are unprefixed (`/dashboard`, `/sign-in`, …), but every real URL is locale-prefixed (`/en/dashboard`, `/de/sign-in`, …) via `middleware.ts` + `app/[lang]/`. `Disallow: /dashboard` never matched, so private pages were effectively crawlable.
- `app/robots.ts` now expands each non-API disallow path to its `/{locale}{path}` variants for `en|de|fr|it`. `/api/*` paths stay unprefixed since the middleware matcher excludes them from locale routing.
- Added two test cases covering the locale-prefixed variants and the API non-prefixing behaviour.

## Test plan
- [x] `pnpm exec vitest run app/__tests__/robots.test.ts` — 8/8 passing
- [ ] Sanity-check `/robots.txt` on the preview deploy contains `/en/dashboard`, `/de/sign-in`, etc., and still contains `/api/auth/` without a locale prefix.

### DIFF
--- a/apps/web/app/__tests__/robots.test.ts
+++ b/apps/web/app/__tests__/robots.test.ts
@@ -32,6 +32,19 @@ describe("robots", () => {
     expect(disallow).toContain("/sign-up");
   });
 
+  it("disallows locale-prefixed variants of private pages", () => {
+    const result = robots();
+    const rules = Array.isArray(result.rules) ? result.rules : [result.rules];
+    const wildcard = rules.find((r) => r.userAgent === "*");
+    const disallow = wildcard!.disallow as string[];
+    for (const locale of ["en", "de", "fr", "it"]) {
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Critic approved. Two minor follow-ups flagged: (1) the bare `/dashboard` entry is dead weight after locale-expansion because middleware redirects unprefixed requests — PR body already calls this out; (2) worth adding a code comment clarifying that the `(app)` route group contains both public (`/company`, `/explore`, `/watchlists`, `/[userSlug]`) and private (`/dashboard`, `/settings`, `/my-jobs`, `/progress`) routes, to prevent a future contributor from blanket-disallowing `(app)/*`.

PR #2236, opened 2026-04-19
Fix IndexNow watchlist notify silently broken (after() outside request scope)

viktor-shcherb:
## Summary

The web-side IndexNow notifier silently stopped firing for watchlist mutations.

**Symptom:** Bing's IndexNow dashboard shows the most recent submission at 17 Apr 2026 14:00 UTC (the crawler-side bootstrap of 16,684 company-locale URLs). Since then, **11 public watchlists were created** between Apr 13 and Apr 17 — including 2 *after* the bootstrap (16:51 and 18:38 UTC on Apr 17) — and **none reached IndexNow**.

The crawler-side container is healthy: hourly ticks all log `indexnow.nothing_to_submit unchanged=16684`, which is the expected steady state when no company stable fields or descriptions have changed in Postgres (`companies.csv` last touched 2026-04-09). The gap is web-side only.

## Root cause

`notifyIndexNow` (`apps/web/src/lib/indexnow.ts`) wrapped its `fetch` in `after()` from `next/server`. But every call site in `apps/web/src/lib/actions/watchlists.ts` invoked it from inside a detached promise:

```ts
_getOwnerInfo(userId).then((owner) => {
  // ...
  notifyIndexNow([`/${userSlug}/${slug}`]);  // calls after() here
}).catch((err) => { console.error(err) });

return { id, slug };
```

`after()` must be called **synchronously inside the request scope**. Once the action returns and `workAsyncStorage` exits, the inner `after()` throws `"after() was called outside a request scope"`. The wrapping `.catch()` swallowed it. (And once the request is torn down, even the `console.error` itself may not reach Vercel's log sink — which explains why a `vercel logs --since=6d --level error` grep across 5,000+ lines for `request scope` / `hook failed` / `indexnow` returns zero matches. The same lifecycle that breaks `after()` also drops late `console.error`.)

## Fix

- **`apps/web/src/lib/indexnow.ts`** — `notifyIndexNow` becomes a plain async function. Awaits its fetch directly; no internal `after()`. Docstring documents that the caller in a Vercel handler should wrap in `after()`.
- **`apps/web/src/lib/actions/watchlists.ts`** — wraps the post-mutation hook in `after(async () => { … })` at four call sites: `createWatchlist`, `updateWatchlist` (both indexed + unindex paths), `deleteWatchlist`, `copyWatchlist`. The Typesense upsert/delete invocations were moved inside the same `after()` block for consistency.
- **`apps/web/src/lib/__tests__/indexnow.test.ts`** — new file, 11 tests covering: no-op paths, locale expansion (exact URL list per path), encoding edge cases, dedupe + ordering, 10,000 URL cap, success-is-silent, 4xx error log includes status, network error logging, structural guard that the file does **not** import from `"next/server"`, and that the function awaits its fetch.
- **`docs/13-seo-and-indexnow.md`** — updated the implementation note to reflect that callers own the `after()` wrap and call out why.

## Known limitations / out of scope

- **Typesense fire-and-forget**. `tsUpsertWatchlist` / `tsDeleteWatchlist` / `tsUpdateWatchlistField` (in `apps/web/src/lib/search/typesense-watchlist.ts`) still return `void` and detach their internal `.upsert(...).catch(...)` promise. Wrapping their *invocations* in `after()` keeps the function alive long enough for the call to be issued, but doesn't await its completion. Making Typesense reliably durable would require changing those helpers to return `Promise<void>` — separate PR.
- **`addCompanyToWatchlist` / `clearWatchlistCompanies` / `removeCompanyFromWatchlist`** (~lines 1031, 1066) have the same detached `_countWatchlistCompanies(...).then(tsUpdateWatchlistField)` shape but it's Typesense-only with no IndexNow involvement, so they're not in this PR's blast radius.
- **Action-level integration tests**. There are no tests for `apps/web/src/lib/actions/watchlists.ts` (no `actions/__tests__/` directory exists). A useful follow-up is one test per action that mocks `next/server`'s `after`, captures the registered callback, runs it, and asserts `notifyIndexNow` is invoked with the expected URL list. Heavy mock surface (db / session / typesense / indexnow) so deferred to its own PR.

## Test plan

- [x] `pnpm test` in `apps/web` — 162 tests pass (11 new in `indexnow.test.ts`)
- [x] `pnpm tsc --noEmit` — clean (the only error is pre-existing `@jseek/mcp-server/handler` not built locally)
- [x] Verified `INDEXNOW_KEY` is set on Vercel: `curl https://jseek.co/indexnow-key.txt` returns `70974894bca8f34855a86ba0efd14d54` (matches the crawler's env)
- [ ] After deploy: create a public watchlist on prod and confirm a new entry appears on Bing's IndexNow submission dashboard within a few minutes

## Verification details (kept here for the postmortem)

- Crawler container: `deploy-indexnow-1` up 39h, hourly ticks, all `unchanged=16684`
- Postgres `indexnow_submission`: 16,684 rows, all `last_submitted_at` between `2026-04-17 12:00:17` and `12:00:19` UTC (the v2 hash bootstrap window)
- Supabase `watchlist`: 11 public rows since 2026-04-13, none reflected in Bing dashboard
- Vercel logs: 0 hits across 5,000+ lines for `request scope` / `hook failed` / `indexnow` — consistent with the `console.error` from the detached `.catch()` itself being dropped after the function exits

### DIFF
--- a/apps/web/src/lib/__tests__/indexnow.test.ts
+++ b/apps/web/src/lib/__tests__/indexnow.test.ts
@@ -1,3 +1,6 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { notifyIndexNow } from "../indexnow";
@@ -60,11 +63,33 @@ describe("notifyIndexNow", () => {
     );
   });
 
-  it("dedupes URLs across paths", async () => {
+  it("dedupes URLs across paths and pins ordering", async () => {
     process.env.INDEXNOW_KEY = "test-key";
     await notifyIndexNow(["/foo", "/foo"]);
     const body = JSON.parse(fetchMock.mock.calls[0][1].body as string);
-    expect(body.urlList).toHaveLength(4); // 4 locales × 1 unique path
+    expect(body.urlList).toEqual([
+      "https://jseek.co/en/foo",
+      "https://jseek.co/de/foo",
+      "https://jseek.co/fr/foo",
+      "https://jseek.co/it/foo",
+    ]);
+  });
+
+  it("caps the batch at 10_000 URLs (IndexNow protocol limit)", async () => {
+    process.env.INDEXNOW_KEY = "test-key";
+    // 2_501 paths × 4 locales = 10_004 expanded URLs → must be trimmed.
+    const paths = Array.from({ length: 2_501 }, (_, i) => `/p${i}`);
+    await notifyIndexNow(paths);
+    expect(fetchMock).toHaveBeenCalledOnce();
+    const body = JSON.parse(fetchMock.mock.calls[0][1].body as string);
+    expect(body.urlList).toHaveLength(10_000);
+  });
+
+  it("does not log on a 200 response (success is silent)", async () => {
+    process.env.INDEXNOW_KEY = "test-key";
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    await notifyIndexNow(["/foo"]);
+    expect(errSpy).not.toHaveBeenCalled();
   });
 
   it("logs (does not throw) when the endpoint rejects with 4xx", async () => {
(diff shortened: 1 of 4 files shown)


viktor-shcherb:
Critic approved. One follow-up flagged: `tsUpsertWatchlist` / `tsDeleteWatchlist` / `tsUpdateWatchlistField` are still sync `void` functions that detach internally; wrapping them in `after()` keeps the host alive long enough to *issue* the Typesense request but doesn't extend to completion. Worth making them return `Promise<void>` in a follow-up so `after()` actually awaits the fetch. PR body already notes this as explicit scope exclusion.

PR #2240, opened 2026-04-19
Route Uber scraper through Webshare proxy

viktor-shcherb:
## Summary

Sets `scraper_config.proxy: true` on `uber-careers`. `www.uber.com` has been serving a Cloudflare challenge page (`def.uber.com/en/challenge`) to Hetzner egress IPs, so the json-ld scraper never sees a `JobPosting` — **567 recent scrapes, 97% silent failures**. Webshare clears the WAF cleanly.

## Root cause

`c204b210` (merged 2026-04-17) fixed Uber's 406 rejection by sending a browser `Accept` header. That turned 406s into 200s — but the 200 is a CF-managed challenge page, not the job detail. The json-ld extractor finds no `<script type="application/ld+json">` and emits silent `jsonld.not_found`.

The redirect chain from Hetzner:

```
/careers/list/{id}
  → 301 /global/en/careers/list/{id}/
  → 200 https://def.uber.com/en/challenge?from_service=... (35 KB HTML, no JSON-LD)
```

## Verified from Hetzner production image

`ghcr.io/colophon-group/jobseek-crawler-browser:latest` against 10 live-from-monitor URLs:

```
# Current (no proxy)
status=200 jsonld=0  final=https://def.uber.com/en/challenge?from_service=...   (×10)

# With webshare (this PR)
status=200 jsonld=1  final=https://www.uber.com/global/en/careers/list/157754/
status=404                                                                        (gone)
status=200 jsonld=1  final=https://www.uber.com/global/en/careers/list/158001/
status=404                                                                        (gone)
... (etc)
```

End-to-end through `jsonld.scrape()` + `create_http_client(use_proxy=True)`:

```
OK title='Account Executive, SMB Acquisition, UberEats ANZ'  locs=['Sydney, NSW, AUS']
HTTPStatusError 404
OK title='Strategy and Planning Associate'  locs=['Toronto, Ontario, CAN']
HTTPStatusError 404
HTTPStatusError 404
HTTPStatusError 404
HTTPStatusError 404
HTTPStatusError 404
OK title='Software Engineer II, Android - Family & Teens'  locs=['Rio de Janeiro, …, BRA']
HTTPStatusError 404

ok=3  gone=7  other=0
```

The 3/10 live postings extract full titles + locations. The 7/10 are **genuinely gone** (404, which the existing scrape pipeline already handles via the 3-strike `_RECORD_SCRAPE_FAILURE` cycle).

## Why not TLS impersonation / header tuning?

[...]

PR #2242, opened 2026-04-19
Treat Workday 403 + errorCode=S22 as detail-gone (peels #2214)

viktor-shcherb:
## Summary

Extend the Workday detail-scraper's existing 404 "posting gone" handler to also catch the Workday-specific **403 with `{errorCode: "S22"}`** shape. That's the response Workday's CXS endpoint returns for requisitions that were closed/unlisted between monitor-tick and scrape-tick — and it's responsible for the bulk of the WAF-flavoured errors tracked in #2214.

Peels a large slice off #2214 without needing a proxy or residential egress.

## Investigation trail (all from Hetzner's production IP)

1. **Top 403 host for 24h: `wf.wd1.myworkdayjobs.com` (Wells Fargo), 1443 errors.** 14 other `.wd*.myworkdayjobs.com` tenants together add ~3,500 more.
2. Scanned 29 random Workday tenants' detail endpoints — **28 succeed** on first GET. So this isn't a blanket Workday-on-Hetzner block.
3. Tested 10 repeated GETs on the highest-403 tenants including Wells Fargo — **10/10 success** on the FIRST posting returned by LIST.
4. Tested the **exact URLs from Loki's last 2h of 403s** — 15/15 still 403, all with:

```json
{"errorCode":"S22","errorCaseId":"…","httpStatus":403,"message":"permission denied","messageParams":{}}
```

5. Pulled the current full WF listing (`POST /wday/cxs/wf/WellsFargoJobs/jobs`) — **0/14 of the 403-URLs are in the current listing.** They are all genuinely delisted; Workday returns 403+S22 instead of 404.

6. Playwright + full cookie warmup (`page.goto(html) → ctx.request.get(cxs)`) ALSO returns 403+S22 for these URLs. Confirming this isn't a TLS/fingerprint issue — it's the app-level "this requisition isn't published anymore" response.

## Fix

```python
if resp.status_code == 404:
    log.info("workday_scraper.detail_gone", url=url, status=404)
    return JobContent()
if resp.status_code == 403 and _is_gone_response(resp):
    log.info("workday_scraper.detail_gone", url=url, status=403, error_code="S22")
    return JobContent()
resp.raise_for_status()
```

`_is_gone_response` parses the body and checks `errorCode == "S22"`. Anything else (non-JSON, different errorCode, no body) still raises so real auth failures / WAF blocks / rate limits keep surfacing in `batch.scrape.error`.

## Tests

Four new cases pin the branching so a future refactor doesn't silently swallow something else:

| case | expected |
|---|---|
| 403 + `{errorCode:"S22"}` | empty JobContent, no exception |
| 403 + `{errorCode:"OTHER"}` | raises `HTTPStatusError` (not gone) |
| 403 with non-JSON body | raises `HTTPStatusError` (real WAF block) |
| bare 403 | raises `HTTPStatusError` (existing behaviour) |

`pytest tests/test_workday.py` — 60 passed.

## Expected impact

From today's Loki:
- 14 tested tenants: **15/15 of today's 403s are errorCode=S22 gone-signals**.
- ~3,500 Workday 403 events/24h expected to disappear from `batch.scrape.error` and move to `info`-level `workday_scraper.detail_gone`.
- 3-strike failure cycle no longer churning on delisted requisitions; scrape queue drains cleanly once the monitor's next tick unsets `is_active`.

## What this does NOT fix

The minority of 403s that ARE real WAF blocks (non-JSON responses, HTML challenge pages) will still surface. Those remain in #2214's scope — they need the proxy/browser approach (Uber's fix in #2240 is the template; Workday's ecosystem needs a different proxy tier that its WAF doesn't ban).

## Test plan

- [x] `pytest tests/test_workday.py` — 60 passed
- [x] Live Hetzner probe: 15/15 matching URLs, 0/14 in current LIST
- [ ] After deploy: `batch.scrape.error{host=~"*.myworkdayjobs.com"}` drops 70-90%; `workday_scraper.detail_gone` at equivalent info-level volume


viktor-shcherb:
One correction from review: the PR body said the 3-strike `_RECORD_SCRAPE_FAILURE` cycle would keep draining these. That's wrong — the empty-JobContent path goes through `_RECORD_SCRAPE_SUCCESS`, which resets `scrape_failures = 0`. So 403+S22 becomes a **soft-success** (same as the existing 404 → empty case), not a soft-fail. Net effect is the same (scrape queue drains cleanly) but the mechanism is cleaner than I first wrote — the monitor's gone-detection is what eventually unsets `is_active`, no retry churn on our side.

Follow-ups noted by the critic, not blocking:
- Only Wells Fargo was empirically verified. S22 semantics should match across tenants (it's a platform-level code) but post-deploy worth sampling Target/HPE/Citi/MSFT/etc. via `workday_scraper.detail_gone` volume vs. LIST membership.
- Consider adding an `error_code` label on the raising path so the residual ~10-30% of 403s has a histogram showing what's in it — informs #2214 triage.

PR #2445, opened 2026-04-22
Set wefox.name to unblock deploys (prod is down)

viktor-shcherb:
## Summary

`wefox` was added in #2427 with all fields NULL. `crawler sync` upserts companies.csv into the `company` table, which has `name TEXT NOT NULL`. Every deploy since 07:46 UTC has failed with:

```
asyncpg.exceptions.NotNullViolationError: null value in column "name" of relation "company" violates not-null constraint
DETAIL: Failing row contains (c15bd2ec..., wefox, null, null, ...).
```

All 8 deploys since 07:46 failed at the sync step, so the Hetzner compose stack is DOWN — only redis is running. No workers, browser-1, exporter, drain, or indexnow.

This PR populates `wefox.name = "Wefox"` + website, which is the minimum to pass the NOT NULL constraint. No board is configured for wefox yet, so the company is a no-op until a monitor is added later.

## Follow-ups (not in this PR)
- Harden `sync.py` to skip + warn on NULL-name rows so a broken stub CSV can never brick all deploys again.
- Revisit the company-request workflow so stubs aren't merged to main before having at least a name.

## Test plan
- [ ] Merge → next deploy run should pass the sync step
- [ ] Verify `docker compose ps` on browser worker shows all services healthy after deploy completes


viktor-shcherb:
Superseded by #2446 (broader cleanup removing all 14 orphan companies).

PR #2471, opened 2026-04-23
Fix sitemap monitor for https:// XML namespace variant

viktor-shcherb:
## Summary

- Some career portals (e.g. TalentsConnect Job Shop used by NEURA Robotics) emit `xmlns="https://www.sitemaps.org/schemas/sitemap/0.9"` (HTTPS) instead of the canonical `http://` in their sitemap XML
- Python's `xml.etree.ElementTree` treats `https://` and `http://` as distinct namespaces, so `_extract_urls()` found no `<url>` elements and returned 0 results despite the sitemap containing valid job URLs
- Added `_detect_ns()` helper that extracts the actual namespace prefix from the root element tag at parse time
- `_extract_urls()` and `_extract_child_sitemaps()` now use the detected prefix instead of the hardcoded `http://` constant
- The no-namespace fallback path is retained for fully namespace-free sitemaps
- Added tests: `test_https_namespace_variant` and `TestDetectNs` class (3 cases)

## What was tried first

- Configured `sitemap_url` explicitly in the board config — still 0 results
- Tried `url_transform` and `url_filter` variations — still 0 results
- `curl` confirmed the sitemap has 149 `/offer/` URLs but the parser returned 0

## Test plan

- [ ] `uv run pytest tests/test_sitemap.py` — all 40 tests pass (7 new)
- [ ] Verify NEURA Robotics sitemap now returns 129 jobs via `ws run monitor`

### DIFF
--- a/apps/crawler/src/core/monitors/sitemap.py
+++ b/apps/crawler/src/core/monitors/sitemap.py
@@ -14,6 +14,8 @@ log = structlog.get_logger()
 
 MAX_URLS = 50_000
 NS = "{http://www.sitemaps.org/schemas/sitemap/0.9}"
+# Some generators emit https:// instead of http:// in the namespace declaration.
+NS_HTTPS = "{https://www.sitemaps.org/schemas/sitemap/0.9}"
 
 _JOB_KEYWORDS = ("job", "career", "posting", "position", "vacancy", "opening")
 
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Leaving for human code review (labeled `review-code`). Orchestrator does not auto-merge code fixes.

PR #2438, opened 2026-04-22
Add Jobylon monitor type

viktor-shcherb:
Closes #1955

## Orakl Oncology
https://www.orakl-oncology.com
logo_type: wordmark+icon

| Full Logo | Minified Logo |
|-----------|----------------|
| full-logo | minified-logo |

|  | orakl-oncology-wttj |
|---|---|
| URL | https://www.welcometothejungle.com/en/companies/orakl-oncology/jobs |
| Monitor | `dom` · `{"render": false}` |
| Scraper | `json-ld` |
| Jobs | 16 |
| Cost | ~0.46s/cycle |
| title | 3/3 (clean) |
| description | 3/3 (clean) |
| locations | 3/3 (clean) |
| employment_type | 3/3 (clean) |
| job_location_type | 0/3 (absent) |
| date_posted | 3/3 (clean) |
| base_salary | 0/3 (absent) |
| skills | 0/3 (absent) |
| qualifications | 3/3 (clean) |
| responsibilities | 0/3 (absent) |
| valid_through | 3/3 (clean) |
| **Verdict** | **good** — DOM monitor finds 16 jobs from WTTJ. JSON-LD extracts all required fields cleanly. Employment_type normalized to full_time. Date_posted, qualifications, valid_through also extracted. job_location_type absent from JSON-LD. |

PR #2495, opened 2026-04-23
Jobylon parser hardening + transport-error semantics

viktor-shcherb:
## Why

Three review findings on #2438 — two parser blockers and one error-handling bug that would cause spurious board delistings. Stacked on #2438's branch so it can be squashed in before merge.

### B1 — ``_JOBS_BLOCK_RE`` truncates at ``];`` inside a string

The capture ``(\[.*?\]);`` with non-greedy ``.*?`` stops at the *first* ``];`` — including ones inside quoted values. Verified empirically:

```python
>>> html = \"JBL.embed_v2['jobs'] = [{id: '1', title: 'lead with ];'}, {id: '2'}];\"
>>> _JOBS_BLOCK_RE.search(html).group(1)
\"[{id: '1', title: 'lead with ]\"  # truncated at the first ']'
```

Any posting with ``];`` in title/summary/locations_text silently drops the rest. Fix: anchor on the ``JBL.embed_v2['jobs'] = [`` prefix, then scan for the matching ``]`` with a string-aware bracket counter that tracks single/double quotes + ``\\`` escapes.

### B2 — unquoted-key rewriter mutates strings

Order was: key-quote pass → string rewrite. The key regex ``[\{,]\s*[A-Za-z_]\w*:`` doesn't know about string boundaries, so a value like:

```
summary: 'config snippet: {host: localhost, port: 80}'
```

became:

```
summary: 'config snippet: {\"host\": localhost, \"port\": 80}'
```

Fix: convert strings to double quotes first, then run a dedicated ``_quote_unquoted_keys_outside_strings`` that tracks in-string state and only substitutes outside.

### N2 — transport errors treated as ``BoardGoneError``

``_fetch_embed`` caught ``httpx.HTTPError`` → returned ``None`` → ``discover`` raised ``BoardGoneError``. One ``ConnectError`` during a Cloudflare hiccup flips a healthy tenant to ""gone"" and starts delist processing. Only 404 is a permanent-gone signal now; transport errors bubble up and get retried by the worker.

## Verification

- 4 new unit tests: B1 (string ``];``), B2 (string ``{foo:``), unterminated-array safety, transport-error propagation.
- The pre-existing ``test_transport_error_raises_board_gone`` test asserted the bug as intentional; inverted to ``test_transport_error_propagates``.
- Full suite: ``uv run pytest tests/`` — **3321 passed**.

## Merge order

Squash into #2438 at merge time, or merge this on top after #2438.

### DIFF
--- a/apps/crawler/src/core/monitors/jobylon.py
+++ b/apps/crawler/src/core/monitors/jobylon.py
@@ -65,13 +65,13 @@ _JOBYLON_HOSTS: frozenset[str] = frozenset(
 # ``job-lang-XX``.  We prefer the latter.
 _LANG_CODE_RE = re.compile(r"job-lang-([a-z]{2})")
 
-# Top-level embed-jobs block we need to capture.  The delimiter uses
-# single quotes inside a bracketed key to avoid confusion with the
-# widget's ng-init attributes.
-_JOBS_BLOCK_RE = re.compile(
-    r"JBL\.embed_v2\['jobs'\]\s*=\s*(\[.*?\]);",
-    re.DOTALL,
-)
+# Anchor for the top-level embed-jobs assignment.  We locate the
+# ``JBL.embed_v2['jobs'] = [`` prefix via regex, then use a
+# string-aware bracket scanner to find the matching ``]``.  A pure
+# ``(\[.*?\]);`` capture would terminate at the first ``];`` that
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Superseded by #2503

Issue #2641, opened 2026-04-25
Fluid: cache or move parseSearchFilters' N×4 Postgres suggest fanout


viktor-shcherb:
## Context
\`parseSearchFilters\` is called from every search server action (\`fetchExploreData\`, \`fetchCompanyPageData\`, \`/api/v1/search\`). When the user has any keywords, it runs an expensive parallel fanout of Postgres suggest queries that scales with token count.

## Evidence
\`apps/web/src/lib/actions/search-input.ts:244-263\`:

\`\`\`ts
const [senResults, locResults, occResults, techResults] = await Promise.all([
  Promise.all(singles.map((c) => suggestSeniorities({ query: c, locale }))),
  Promise.all(singles.map((c) => suggestLocations({ query: c, locale, userLat, userLng }))),
  Promise.all(allCandidates.map((c) => suggestOccupations({ query: c, locale }))),
  Promise.all(singles.map((c) => suggestTechnologies({ query: c, locale }))),
]);
\`\`\`

For a 5-keyword query: 5 senior + 5 location + ~14 occupation candidates (5 singles + 4 pairs + 3 triplets, capped at length≤10) + 5 tech = **~29 parallel Postgres queries** per server action. \`suggestLocations\` uses a recursive CTE per the comment at line 243.

This runs on every \`searchJobs()\`/\`listTopCompanies()\` call from \`/api/v1/search\` and the page-load actions in #2640.

## Proposal
Two options, in priority order:

1. **Cache per-(token, locale) suggestion results in Redis**, TTL 1h. Common terms ("python", "engineer", "berlin") will hit cache with very high rate. Wrap each \`suggest*\` call in \`cached(...)\`. Estimated ~80%+ hit rate after warmup.

2. **Move parsing client-side** using a small precomputed dictionary of canonical occupations/seniorities/technologies (already shipped to clients via \`/api/v1/taxonomies\`). Locations stay server-side because of the recursive CTE + geo bias. This stacks with #2641.

Either way, cap occupation candidate fanout: words.length≤10 currently spawns up to 8 triplet candidates per word — bound the candidate set tighter.

## Expected impact
Removes ~10–30 Postgres round-trips per keyword search. Each saves ~10–30ms of function time even when parallelized (postgres.js connection pool serializes when \`max: 10\` is saturated). High-volume win.

Issue #2640, opened 2026-04-25
Fluid: server-render initial data for explore/company/watchlist instead of client-mount fetch


viktor-shcherb:
## Context
The three highest-traffic dynamic pages all use the same anti-pattern: server page renders a skeleton, client \`useEffect\` fires a server action on mount to fetch data. Every page visit (including signed-out humans and SEO crawlers that execute JS) burns Vercel function time before users see content.

## Evidence
- \`apps/web/app/[lang]/(app)/explore/explore-content.tsx:20-27\` — \`useEffect(() => { fetchExploreData(...).then(setData) }, [])\`
- \`apps/web/app/[lang]/(app)/company/[slug]/company-content.tsx:30-39\` — same pattern, \`fetchCompanyPageData\`
- \`apps/web/app/[lang]/(app)/[userSlug]/[watchlistSlug]/watchlist-content.tsx:26-32\` — same pattern, \`fetchWatchlistPageData\`

Each of those server actions runs ~80–350ms of work per the doc (\`docs/edge-requests.md\` §"Per-route compute profile"). The \`/explore\` and \`/company/[slug]\` shells themselves don't read request data and are eligible for ISR.

## Proposal
For the unauthenticated, no-filter case (the bulk of organic traffic):
1. Move the initial data fetch into the page server component, fetch defaults via ISR (revalidate=60 for explore, =600 for company), and pass to the client component as \`initialData\`.
2. Client only refetches when the user actually changes filters — and per #2641, that path becomes browser→Typesense direct.

This pairs with #2641 (direct Typesense) — together they eliminate both initial-mount AND per-filter Vercel function calls for /explore and /company/[slug].

## Constraints
- Page must stay statically prerenderable. Watch the ISR test guard at \`apps/web/app/__tests__/isr-routes.test.ts\` (no \`searchParams\`/\`headers()\`/\`cookies()\` reads on the server render path).
- Personalized data (\`getPreferences\`, \`getViewerLanguages\`, geolocation from \`getGeoFromHeaders\`) cannot ride along with ISR. Solutions: (a) accept default currency/language for the prerendered first paint, hydrate personalized version client-side post-mount; (b) keep the fetch only for logged-in users (gate by the existing \`logged_in\` cookie hint).

## Expected impact
Cuts initial-mount function invocations roughly in proportion to anonymous traffic share. SEO crawlers already see static HTML (good) but humans currently always trigger a function — even one repeat visit per minute per active user adds up.

[...]

PR #2746, opened 2026-04-26
Server-render /explore initial data for ISR-eligible homepage (partial #2640)

viktor-shcherb:
## Summary

Partial fix for #2640: `/explore` becomes ISR-eligible and serves prerendered initial data for anonymous, no-filter visitors. The other two pages from #2640 (`/company/[slug]`, `/[userSlug]/[watchlistSlug]`) are explicit follow-ups.

## Why

`page.tsx` previously declared `searchParams: Promise<...>` in `Props` and `ExploreContent` fired `fetchExploreData` from `useEffect` on every mount. Both opt the route into dynamic rendering — every visit (including SEO crawlers + humans alike) pays a Vercel function invocation, even when the response is the locale's default no-filter shape that's identical for every anonymous visitor.

## What changed

1. **`page.tsx`** declares `revalidate = 60`, drops `searchParams` from Props, and calls a new `fetchExploreDefaults({ locale })` that returns the no-filter, anonymous-default `ExploreData` — no `getPreferences` (cookies), no `getGeoFromHeaders` (headers). Result embedded in prerendered HTML as `initialData`.
2. **`ExploreContent`** accepts optional `initialData`. On mount it conditionally re-fetches the personalized variant via `fetchExploreData` only when (a) the user has the `logged_in` hint cookie OR (b) a filter searchParam is present (`q/loc/occ/sen/tech/sal/salcur/exp`). Anonymous, no-filter visitors get zero function invocations.
3. **`listTopCompanies` refactor** — extracted session-free `_listTopCompaniesImpl` core + a thin session-reading `listTopCompanies` wrapper + new `listTopCompaniesAnonymous`. The original called `getSessionUserId()` → `await headers()`, which would silently downgrade `/explore` to dynamic. **Caught by cold-read review** — would have defeated the whole optimization if shipped.
4. **ISR test guard expanded** — `/explore` added to `ISR_PROTECTED_PAGES`. `fetchExploreData` and `listTopCompanies` added to `TAINTED_HELPERS` so a future regression in `page.tsx` calling the personalized / session-reading variant trips the test. Word-boundary regex correctly distinguishes `fetchExploreDefaults(` (safe) from `fetchExploreData(` (tainted), and `listTopCompaniesAnonymous(` (safe) from `listTopCompanies(` (tainted) — verified empirically by round-2 critic.

## Personalization trade-off

Per the issue's solution (a): logged-in users see the anonymous-default first paint (EUR currency, no language filter, no geo bias) for ~50ms before `useEffect` fires and `setData` re-renders. Acceptable per the issue spec — anonymous traffic is the bulk of organic visits, and the win there is zero-function CDN serving.

## Out of scope (follow-ups against #2640)

- `/company/[slug]` — already in `ISR_PROTECTED_PAGES` but `company-content.tsx` still mount-fetches.
- `/[userSlug]/[watchlistSlug]` — same story.
- Companion direct-Typesense work (#2641) is not coupled here.

## Test plan

- [x] 6 new `ExploreContent` unit tests: anonymous-no-filter (no fetch), `logged_in` hint (fetch), each of 8 filter searchParams (fetch via parametric loop), legacy `initialData=undefined` (fetch), utm-non-filter (no fetch).
- [x] `isr-routes.test.ts` covers `/explore` page module + co-located server siblings.
- [x] Related tests pass: client-cookies, search, ISR (25/25).
- [ ] Vercel preview smoke check: `curl -sSI https://<preview>/en/explore` expects `cache-control: public` and (after warmup) `x-vercel-cache: HIT`. Document this in the PR's deploy checklist before merge.


viktor-shcherb:
Reopen to retrigger CI

Issue #2758, opened 2026-04-28
J4: SSRF allowlist for board_url / sample_url


viktor-shcherb:
## Goal
Shared util used by every subcommand route that accepts a URL from the agent. Allowlist by host pattern, with DNS-rebinding protection.

## References
- DESIGN.md §3.6 (Publisher SSRF defense), §4.2 (Probe SSRF defense)

## Behavior
- Allowlist patterns (initial set; expand as needed):
  - `*.greenhouse.io`, `boards.greenhouse.io`, `job-boards.greenhouse.io`
  - `jobs.lever.co/*`, `*.lever.co`
  - `*.myworkdayjobs.com`
  - (full list to be drawn from `apps/crawler/data/boards.csv` patterns)
- Resolve DNS once. Reject if resolved IP is private (RFC1918), loopback, link-local, or matches metadata-service IPs (`169.254.169.254` etc.)
- Re-check at fetch time using a custom DNS-rebinding-aware fetcher (resolve, post-resolution, fetch the IP directly with the original `Host` header)

## Verification

**Tests** (`apps/web/src/lib/murmur/ssrf.test.ts`):
- Allowed host (`*.greenhouse.io` match) + public IP → passes
- Allowed host + private IP (mocked DNS resolves to `192.168.1.1`) → rejected with `url_resolves_to_private`
- Allowed host + metadata IP (`169.254.169.254`) → rejected
- Disallowed host → rejected with `url_not_allowed` (no DNS lookup performed)
- Loopback (`localhost`, `127.0.0.1`) → rejected
- Link-local (`169.254.0.0/16`) → rejected
- IPv6 ULA (`fc00::/7`) → rejected
- DNS rebinding scenario: first lookup resolves public, second lookup resolves private; fetch uses post-resolution IP and rejects → rejected
- Test patterns are documented in source as a static list

**Quality gates:**
- Module exports a single `validateUrl(url: string): { ok: true } | { ok: false, error: string }` function
- All callers of `validateUrl` are at route boundaries; no use deeper in the call stack
- Patterns documented in source comments; updates require a code change (no env-var override)

## Definition of done
- [ ] Tests cover all 8 cases above
- [ ] Module used in J5/#2759's routes (verified via codepath grep)

**Parent:** #2753

viktor-shcherb:
Claimed by orchestrator-developer until 2026-04-29T11:44:46Z.

viktor-shcherb:
## Implementation sketch

[...]

PR #2764, opened 2026-04-29
feat(jobseek): per-claim KV table for named configs

viktor-shcherb:
## Summary
Adds the `murmur_claim_kv` Postgres table (composite PK `(claim_token, name)`, jsonb value, `created_at` / `updated_at` timestamps) and a TS module at `apps/web/src/lib/murmur/claim-kv.ts` exposing `getKV` / `setKV` / `listKV` / `clearKV` for jobseek's named-config state. `setKV` is an UPSERT keyed on the composite PK; concurrent writers are last-write-wins via Postgres row-level locking.

## Related issue
Closes colophon-group/jobseek#2757

## Files changed (high-level)
- `apps/web/src/db/schema.ts` — adds the `murmurClaimKv` Drizzle table definition (composite PK + `murmur_claim_kv_token_idx` index).
- `apps/web/drizzle/0074_add_murmur_claim_kv.sql` — hand-written migration matching the issue's DDL exactly. Hand-written because the existing drizzle journal is severely out-of-sync with the live schema (snapshots stop at 0028, hand-written migrations run to 0073) and `drizzle-kit generate` prompts interactively for hundreds of unrelated phantom diffs; recent migrations 0072 and 0073 were also hand-written for the same reason. Reconciling the journal is out-of-scope for this issue and would change the diff for every column in the repo.
- `apps/web/src/lib/murmur/claim-kv.ts` — public surface. Imports the shared `db` from `@/db` and the `murmurClaimKv` table from `@/db/schema` (no naked SQL, no separate pool).
- `apps/web/src/lib/murmur/claim-kv.test.ts` — seven verification tests using the repo's existing mock-`@/db` pattern (see `lib/actions/__tests__/bootstrap.test.ts`), backed by an in-memory store that respects the composite PK and UPSERT semantics.

## Verification (from the issue)
- [x] setKV then getKV returns the same value
- [x] setKV overwrite under same (token, name) updates updated_at and overwrites value (UPSERT)
- [x] getKV for unknown token/name returns null
- [x] listKV returns all (name → value) pairs for a given token
- [x] clearKV removes all rows for a token; other tokens unaffected
- [x] Concurrent setKV calls under same (token, name) last-write-wins (no error)
- [x] All values are JSON-serializable (jsonb round-trip, no extra encoding)

## Manual checks run locally
- `pnpm test src/lib/murmur/claim-kv.test.ts` ✓ (7/7 pass)
[...]

PR #2766, opened 2026-04-29
feat(jobseek): SSRF allowlist with DNS-rebinding protection for agent URLs

viktor-shcherb:
## Summary
Adds `apps/web/src/lib/murmur/ssrf.ts`, a single-file module that gates every agent-supplied URL on the Murmur subcommand routes. Two phases: (1) host-pattern allowlist (no DNS, fail closed on miss), (2) DNS resolution + IP filter that rejects loopback, RFC1918, link-local, IPv6 ULA, IANA reserved, and cloud-metadata addresses. A companion `safeFetch` performs the outbound call against the captured IP so a second DNS lookup at connect time has nothing to attack.

## Related issue
Closes colophon-group/jobseek#2758

## Files changed (high-level)
- `apps/web/src/lib/murmur/ssrf.ts` — `validateUrl`, `safeFetch`, `matchHostPattern`, `isPrivateIp`, `HOST_ALLOWLIST`, `SafeFetchError`. Single source of truth for the allowlist; comment block points at `apps/crawler/data/boards.csv` for reference and notes the J5 expansion path.
- `apps/web/src/lib/murmur/ssrf.test.ts` — 29 tests covering all 8 verification cases plus matcher and IP-classifier units. DNS is mocked via a hoist-safe global queue so each scenario is deterministic.
- `apps/web/scripts/grep-validateurl-boundary.sh` — boundary gate. Asserts `validateUrl`/`safeFetch` are imported only from `apps/web/app/api/**/route.ts`, the defining module, or the test file.
- `apps/web/package.json` — adds `grep:validateurl-boundary` script.

- [x] Allowed host (`*.greenhouse.io` match) + public IP -> passes
- [x] Allowed host + private IP -> rejected with `url_resolves_to_private`
- [x] Allowed host + metadata IP (`169.254.169.254`) -> rejected
- [x] Disallowed host -> rejected with `url_not_allowed` (no DNS lookup performed)
- [x] Loopback (`localhost`, `127.0.0.1`) -> rejected
- [x] Link-local (`169.254.0.0/16`) -> rejected
- [x] IPv6 ULA (`fc00::/7`) -> rejected
- [x] DNS rebinding scenario: first lookup public, second lookup private; fetch uses post-resolution IP and rejects -> rejected
- [x] Test patterns documented in source as a static list (`HOST_ALLOWLIST` is a `readonly` const, no env-var override)
- [x] Module exports a single `validateUrl(url: string)` function with the exact return shape from the issue
- [ ] Module used in J5/#2759's routes (verified via codepath grep) — note: J5 is unmerged; this DoD item is intentionally unchecked and will be re-verified when J5 lands.

## Manual checks run locally
- `pnpm --filter @jobseek/web test src/lib/murmur/ssrf.test.ts` -> 29/29 pass
- `pnpm --filter @jobseek/web lint` -> clean
- `pnpm --filter @jobseek/web grep:validateurl-boundary` -> clean
- `pnpm --filter @jobseek/web exec tsc --noEmit` -> no new errors (the 3 unrelated existing errors in `app/mcp/route.ts` and `lib/actions/__tests__/company.test.ts` are pre-existing on `murmur-demo`).

## Notes for reviewer
- **Pinning approach.** `safeFetch` rewrites the URL hostname to the captured IP and preserves the original `Host` header. This is the simplest rebinding mitigation that requires no undici internals; the trade-off (TLS hostname check fails for the IP) is acceptable for the demo against public ATS endpoints. If we extend `safeFetch` to broader transport, swap to a `node:https` Agent with custom `lookup` + `servername` so SNI continues to use the original hostname.
- **Allowlist scope.** Six initial patterns covering Greenhouse, Lever, Workday — the dominant ATS providers in `apps/crawler/data/boards.csv`. Per the orchestrator note, I deliberately did not enumerate every host from `boards.csv`; the comment in `ssrf.ts` flags expansion as J5 routes encounter real rejections.
- **Boundary gate.** The bash gate is intentionally tiny and only checks import sites in `apps/web`. CI doesn't yet wire it up automatically; recommend adding it to whatever `grep:all` aggregator J* milestones eventually create.
- **Pre-existing test failure.** `app/mcp/route.test.ts` fails on `murmur-demo` independently of this PR (handler module is missing/uncompiled). Not addressed here.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -17,7 +17,8 @@
     "optimize:svg": "svgo public/*.svg --multipass",
     "screenshots": "tsx script/capture-screenshots.ts",
     "test": "vitest run",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "grep:validateurl-boundary": "bash scripts/grep-validateurl-boundary.sh"
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.888.0",
--- a/apps/web/scripts/grep-validateurl-boundary.sh
+++ b/apps/web/scripts/grep-validateurl-boundary.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# grep-validateurl-boundary.sh
+#
+# Asserts the caller boundary for `validateUrl` (and `safeFetch`) from
+# `apps/web/src/lib/murmur/ssrf.ts`: those names must only be imported
+# from
+#   - apps/web/app/api/**/route.ts          (route handlers)
+#   - apps/web/src/lib/murmur/ssrf.ts       (defining module — re-exports)
+#   - apps/web/src/lib/murmur/ssrf.test.ts  (its own test)
+#
+# Anywhere else is a boundary violation: SSRF validation must happen at
+# the request boundary so the structured `url_*` error code can be
+# surfaced to the agent in the response envelope. Pushing it deeper
+# means individual call sites won't be checked.
+#
+# Usage:  bash apps/web/scripts/grep-validateurl-boundary.sh
+# Exit:   0 if clean, 1 if any unauthorised import is found.
+#
+# @see colophon-group/jobseek#2758
+
+set -euo pipefail
+
+# Resolve repo root from this script's location so the gate works the
+# same whether invoked from CI, lefthook, or `pnpm grep:*`.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+WEB_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+REPO_ROOT="$(cd "$WEB_ROOT/../.." && pwd)"
+
+cd "$REPO_ROOT"
+
+# Files that import `validateUrl` or `safeFetch` from the ssrf module.
+# Two patterns: relative `./ssrf` or `@/lib/murmur/ssrf`.
+matches=$(
+  grep -rEn \
+    --include='*.ts' --include='*.tsx' \
+    "from ['\"](.*lib/murmur/ssrf|\\./ssrf|\\.\\./ssrf|@/lib/murmur/ssrf)['\"]" \
+    apps/web 2>/dev/null || true
+)
+
+violations=""
+while IFS= read -r line; do
+  [ -z "$line" ] && continue
+  file="${line%%:*}"
+  case "$file" in
+    apps/web/app/api/*/route.ts) ;;
+    apps/web/app/api/*/*/route.ts) ;;
+    apps/web/app/api/*/*/*/route.ts) ;;
+    apps/web/app/api/*/*/*/*/route.ts) ;;
+    apps/web/src/lib/murmur/ssrf.ts) ;;
+    apps/web/src/lib/murmur/ssrf.test.ts) ;;
+    *)
+      violations+="$line"$'\n'
+      ;;
+  esac
+done <<< "$matches"
+
+if [ -n "$violations" ]; then
+  echo "validateUrl/safeFetch must only be imported at route boundaries." >&2
+  echo "Offending imports:" >&2
+  printf '%s' "$violations" >&2
+  exit 1
+fi
+
+echo "grep-validateurl-boundary: clean."
--- a/apps/web/src/lib/murmur/ssrf.test.ts
+++ b/apps/web/src/lib/murmur/ssrf.test.ts
@@ -1,15 +1,25 @@
 /**
  * Tests for ssrf.ts.
  *
- * No real network access; all DNS calls are intercepted via a mocked
- * `node:dns/promises`. Each test queues the (address, family) tuples the
+ * No real network access on the unit suites; all DNS calls are
+ * intercepted via a mocked `node:dns/promises`, and `node:http` /
+ * `node:https` `request` are stubbed so we can assert the request
+ * options (especially `lookup`, `servername`, `hostname`) without
+ * opening a socket. Each test queues the (address, family) tuples the
  * resolver should return, in order. The DNS-rebinding case queues two
  * different answers so the second lookup observably differs from the
  * first.
  *
+ * The bottom suite spins up a real loopback `http.Server` to verify
+ * end-to-end that the captured IP is used for the connection while the
+ * original hostname is preserved on the request line / Host header.
+ *
  * @see colophon-group/jobseek#2758
  */
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import http from "node:http";
+import { Readable } from "node:stream";
+import type { AddressInfo } from "node:net";
 
 // Queue of DNS answers consumed in order by the mocked `lookup`.
 // Stored on globalThis so the hoisted `vi.mock` factory can see them
@@ -67,6 +77,7 @@ import {
   isPrivateIp,
   HOST_ALLOWLIST,
   safeFetch,
+  createPinnedLookup,
 } from "./ssrf";
 
 describe("HOST_ALLOWLIST", () => {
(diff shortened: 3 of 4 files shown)


viktor-shcherb:
## Review: REQUEST CHANGES (posted as comment — `--request-changes` blocked because the reviewer is the author)

Process check passes (claim 09:51:21Z, sketch 09:51:53Z, first commit 09:53:37Z; commit ordering interfaces -> tests -> impl). All 8 issue-verification items have real test bodies (not stubs). Matcher is correct against adversarial inputs I tried (`greenhouse.io.attacker.com`, `evil.com#greenhouse.io`, `EVIL.greenhouse.io.attacker.com`, `xn--greenhouse-evil.io`) — all rejected. Boundary gate fires when I plant a violating import in `apps/web/src/lib/murmur/`. Lint clean, typecheck clean modulo the 3 pre-existing errors on `murmur-demo`, ssrf suite is 29/29 green, full web suite is 305/305 modulo the same pre-existing `app/mcp/route.test.ts` import failure that #2764's review noted.

### Blockers

1. **`safeFetch` is broken for any real HTTPS allowlist target** — `apps/web/src/lib/murmur/ssrf.ts:381-393`. The implementation rewrites `URL.hostname` to the captured IP and tries to preserve the original host via `headers.set('host', v.url.host)`. Two problems:
   - **TLS hostname verification will fail.** Greenhouse/Lever/Workday certs are issued for hostnames, not for IPs. Once you change the URL hostname to `104.19.20.21`, undici's TLS layer sets SNI to that IP, the server returns a cert that doesn't cover `104.19.20.21`, and the handshake aborts with `ERR_TLS_CERT_ALTNAME_INVALID`. This is not a "trade-off" with degraded security — it's a function that cannot complete a single HTTPS request to its allowlist targets.
   - **The `Host` header trick does not work with `fetch`.** The Fetch spec lists `Host` as a forbidden header name; undici either ignores it or throws. Even if TLS were not a problem, the virtual-hosted vhost story is fictional.
   - The tests stub global `fetch` entirely, so the broken outbound path is never exercised. The only assertion about the URL is that it contains the IP — true, but proves nothing about whether the request would succeed.
   - **Fix:** the dev's own comment names the right shape — replace the URL-substitution with a `node:https` Agent (or undici Dispatcher) whose `lookup` override returns the captured IP while the URL keeps the original hostname (so SNI + cert validation use the hostname). Example: `new https.Agent({ lookup: (h, opts, cb) => cb(null, capturedIp, family) })`. Then the test asserts `lookup` is called with the hostname and resolves to the captured IP. Today's `safeFetch` should be marked internal/experimental until the swap lands; otherwise J5 will wire it up, hit confusing TLS errors, and the fix-in-the-moment risks being `rejectUnauthorized: false`.

2. **No outbound-call integration test.** Once the Agent-based fix lands, add at least one test that constructs the Agent and asserts (a) `lookup` returns the captured IP, (b) SNI/`servername` is the original hostname, (c) a second `lookup` call (DNS rebinding) is irrelevant because the Agent caches the captured IP. The current `safeFetch — DNS rebinding scenario` suite asserts the URL contains the IP; a future implementation that uses hostname URL + Agent.lookup pinning will need a different assertion shape and that suite must be updated to match.

### Process / DoD
- DoD second item ("Module used in J5/#2759's routes") is intentionally unchecked because J5 is unmerged. Not blocking.

### Nits (non-blocking)
- `grep-validateurl-boundary.sh` is wired into `pnpm grep:validateurl-boundary` but not into any aggregate `grep:all` or CI workflow. PR body acknowledges this. Recommend wiring it into the umbrella gate when one lands.
- `validateUrl` returns `family: 4 | 6` and the cast at line 317 (`resolved.family === 6 ? 6 : 4`) coerces anything-not-6 to 4. Node's `LookupAddress.family` is already typed `4 | 6`; consider `resolved.family as 4 | 6` with a comment, or trust the type.
- `isPrivateIPv6`'s "anything else `::xxxx` is suspicious — treat as private" branch (line 239) is correct fail-closed behaviour but worth a unit test exercising a non-loopback `::abcd` to lock in the policy.
- The `100.64/10` check covers `b >= 64 && b <= 127`. Consider explicit unit tests for `100.64.0.1` and `100.127.255.254` to prove the range bound.

Once blocker (1) is addressed (safeFetch rewritten to preserve TLS via Agent.lookup), with a corresponding test, request re-review.

viktor-shcherb:
Blocker 1 (TLS-correct IP pinning) and blocker 2 (Agent-level test) fixed in 7391a433. Stack chosen: `node:http(s).request` with a custom `lookup` option (no new dependency added). The request URL keeps the original hostname so SNI / TLS SAN matching is correct against Greenhouse / Lever / Workday certs; the connection-time `lookup` returns the captured IP for the expected hostname only and ENOTFOUND for any other name (supports both `all: false` and `all: true` calling conventions). Added: 5 unit tests for `createPinnedLookup`, a mocked `node:https.request` test asserting `hostname === servername === original hostname` and that the pinned lookup is wired in, and a loopback integration test that uses a real `http.Server` on 127.0.0.1 to confirm the lookup hook actually steers the TCP connection. Also addressed the nits: type-narrowing cleanup at the family assignment, plus `::abcd` IPv6 fail-closed and `100.64/10` CGNAT boundary tests. 39/39 ssrf tests green.

viktor-shcherb:
## Review: APPROVE (round 2)

Both round-1 blockers resolved in 7391a433.

### Blocker 1 — TLS correctness
- ssrf.ts:534 `hostname: url.hostname` (original, not IP)
- ssrf.ts:536 `servername: isHttps ? url.hostname : undefined` — the load-bearing assertion. Without it, SNI would default to whatever hostname is on the URL; with the rewrite to use the original hostname, this is now correct for both SNI and SAN matching against real `*.greenhouse.io` / `*.lever.co` / `*.myworkdayjobs.com` certs.
- ssrf.ts:543 `lookup` is the pinned function; connection-time DNS bypassed.
- createPinnedLookup supports both `all: false` and `all: true` (lines 409-420), so modern `net.connect` works.
- Non-expected hostnames fail closed with `ENOTFOUND` (line 404).

### Blocker 2 — Real agent-level tests
- ssrf.test.ts:476-477 asserts the exact triple-equality on the production code path (`opts.hostname === opts.servername === "acme.greenhouse.io"`).
- ssrf.test.ts:487-501 invokes the captured `lookup` directly and verifies it returns the captured IP for the expected hostname and ENOTFOUND for anything else.
- ssrf.test.ts:560-593 spins up a real `http.Server` on 127.0.0.1, calls `http.request` with `createPinnedLookup` (no mocks of production code), and asserts (a) the Host header preserved the synthetic hostname and (b) `req.socket.remoteAddress` is loopback. This proves the lookup hook is honoured by Node's connection layer.

### Adversarial probes
I temporarily planted two bugs locally:
1. `servername` swapped to a literal wrong hostname → mock test fails at line 477 with a clean assertion message.
2. `lookup` callback returning `127.0.0.1` instead of the captured IP → 4 tests fail (mock test, lookup unit tests).

Both probes fired as expected; coverage is real, not theatre.

### Gates (run locally on the worktree)
- `pnpm --filter @jobseek/web test src/lib/murmur/ssrf.test.ts` → 39/39 pass
- `pnpm --filter @jobseek/web lint` → clean
- `pnpm --filter @jobseek/web exec tsc --noEmit` → no errors in ssrf files (pre-existing errors in `app/mcp/route.ts` and `src/lib/actions/__tests__/company.test.ts` are unrelated and on main)
- `bash apps/web/scripts/grep-validateurl-boundary.sh` → clean

### Process
- Claim at 2026-04-29T09:51:21Z (TTL set)
- Sketch at 09:51:53Z, first commit at 09:53:37Z — sketch precedes code ✓
- Commits ordered: interfaces → tests → impl → fix ✓

### Nits (non-blocking)
- The loopback integration test uses plain HTTP, so it does not exercise SNI/SAN handshakes end-to-end. The mock-based assertion at line 477 covers `servername` correctness, and adding a TLS loopback would require a self-signed cert — legitimate scope cut for M0.
- `createPinnedLookup` casts via `as unknown as LookupFunction` (line 421); the jsdoc justifies this by Node's awkward overloaded callback shape — acceptable.

Issue #2763, opened 2026-04-28
P4: Webhook accept handler


viktor-shcherb:
## Goal
Webhook accept handler that takes `final_output` from Murmur and writes the catalog.

## References
- murmur DESIGN.md §3.6 (Webhook accept-handler contract), §4.2 (Final acceptance)
- murmur#34 (M0) for canonical envelope shape `{ ok, errors?, data? }`

## Behavior
1. Verify `Authorization: Bearer <MURMUR_TOKEN>`
2. Cap body at 5 MB; 413 if larger
3. Read `Idempotency-Key` (the run_id); if a row in catalog already references this `run_id`, return 2xx (idempotent success)
4. Re-run probe validation in-process using J1's lib (defense-in-depth: agent's submitted config is re-tested)
5. Write to local Postgres OR append to `companies.csv` / `boards.csv` (operator decides; same code path with a flag)
6. Return 2xx with `{ ok: true, data: { run_id, applied: true } }` or `{ ok: false, errors: [...] }` per M0 envelope

## Verification

**Tests** (`apps/web/src/app/api/murmur/accept/route.test.ts`):
- Bearer missing → 401
- Bearer wrong → 401
- Body > 5 MB → 413
- Idempotent replay (same run_id, same body) → 2xx with `{ ok: true, data: { applied: false, reason: "already_applied" } }`, only one catalog row written
- Idempotent replay (same run_id, different body) → 2xx with applied:false, second body discarded, log warning (run_id is canonical)
- Re-run probes succeed → catalog written, 2xx with `{ ok: true, data: { applied: true } }`
- Re-run probes fail (config doesn't actually work) → no catalog write, return 200 with `{ ok: false, errors: [...] }` so Murmur doesn't retry forever
- Schema-invalid body → 400 with `{ ok: false, errors: ["validation:<path>:<msg>"] }`

**End-to-end test:**
- Stub Murmur fires a webhook with a realistic `final_output` for a known-good fixture company → catalog row appears in test DB / CSV
- Murmur's one-retry behavior verified: simulate 200 + 200 (idempotency holds) and 503 + 200 (retry succeeds, only one row)

**Quality gates:**
- UNIQUE constraint on `(run_id)` in the catalog write — verified by an integration test that violates it
- Re-run probe takes < 30s; if longer, the route times out gracefully (504)
- All response bodies use the M0 envelope (no `{ accepted: ... }`)

## Definition of done
- [ ] Route + tests committed
[...]

PR #2772, opened 2026-04-29
feat(jobseek): Murmur webhook accept handler

viktor-shcherb:
## Summary

- New `POST /api/murmur/accept` handler for Murmur webhook delivery (DESIGN.md §4.1).
- Bearer auth, 5 MB body cap, `Idempotency-Key` (run_id) dedupe, full schema validation, defense-in-depth probe re-run via the J5 cross-language invoker, and a transactional catalog write with backend switch (Postgres default, CSV via `MURMUR_ACCEPT_TARGET=csv`).

## Related issue

Closes colophon-group/jobseek#2763

## Files changed (high-level)

- `apps/web/app/api/murmur/accept/route.ts` — main pipeline + envelope mapping.
- `apps/web/app/api/murmur/_lib/accept-schema.ts` — vendored `final_output` schema (mirrors `add-company.yaml`).
- `apps/web/app/api/murmur/_lib/accept-validate.ts` — array/object dialect validator.
- `apps/web/app/api/murmur/_lib/idempotency.ts` — canonical-JSON SHA-256 helper.
- `apps/web/app/api/murmur/_lib/accept-pipeline.ts` — body parser w/ 5 MB cap + probe re-run with 30 s wallclock race.
- `apps/web/app/api/murmur/_lib/catalog.ts` — Postgres tx (company + job_board + ledger) + CSV append backend.
- `apps/web/src/db/schema.ts` — new `murmurAcceptLog` table (PRIMARY KEY on run_id).
- `apps/web/drizzle/0075_add_murmur_accept_log.sql` — hand-written migration.
- `apps/web/app/api/murmur/__tests__/accept.test.ts` — 17 unit tests covering the full Verification matrix.
- `apps/web/app/api/murmur/__tests__/accept-validate.test.ts` — 11 schema-validator unit tests.
- `apps/web/app/api/murmur/__tests__/idempotency.test.ts` — 8 canonicalisation/hash tests.
- `apps/web/app/api/murmur/README.md` — append accept-handler section + env vars.

- [x] Bearer missing → 401
- [x] Bearer wrong → 401
- [x] Body > 5 MB → 413 (both buffer-size and Content-Length fast-path)
- [x] Idempotent replay (same run_id, same body) → 2xx with `applied: false, reason: "already_applied"`, only one catalog row
- [x] Idempotent replay (same run_id, different body) → 2xx with `applied: false, reason: "body_mismatch"`, second body discarded, warning logged
- [x] Re-run probes succeed → catalog written, 2xx with `applied: true`
- [x] Re-run probes fail → no catalog write, 200 with `{ ok: false, errors: [...] }` (NOT 5xx)
- [x] Schema-invalid body → 400 with `errors: ["validation:<json-pointer>:<token>"]`
- [x] Re-run probe > 30 s → 504 with `errors: ["probe_timeout"]`
- [x] Murmur 200 + 200 retry simulation: idempotency holds, only one row
- [x] Murmur 503 + 200 retry simulation: first attempt returns soft failure, retry succeeds, only one row
- [x] UNIQUE constraint on `(run_id)` in catalog write — enforced via PRIMARY KEY on `murmur_accept_log.run_id`
- [x] All response bodies use the M0 envelope

## Manual checks run locally

- `pnpm test` — all 489 tests pass (1 pre-existing skip on the gated e2e probe).
- `pnpm lint` — clean.
- `pnpm build` — `/api/murmur/accept` is registered in the route table.
- `pnpm exec tsc --noEmit` — no new errors (2 pre-existing in `src/lib/actions/__tests__/company.test.ts`, unrelated).
- `pnpm grep:validateurl-boundary` — clean.

## Notes for reviewer

- **Idempotency model.** Two layers: an in-process `Map<runId, bodyHash>` and a durable `murmur_accept_log` table. The map exists so unit tests can exercise the full re-fire path without a DB. In production both layers are present — the DB UNIQUE on `run_id` is authoritative; the in-process map serves as a hot cache so re-fires never even reach the DB layer for already-applied runs. The Postgres branch in `catalog.ts` writes the ledger row and the catalog rows in the SAME transaction, so a crash between them never leaves a half-applied state. A `CatalogIdempotencyConflict` thrown by a concurrent first-write race is caught by the route and turned into `applied: false, reason: "already_applied"`.
- **Probe re-run cross-language barrier.** The handler reuses the J5 `invokeLib` so the same Python lib that runs the agent's subcommands is used to re-validate every board's `board_url`. The fan-out is wrapped in a `Promise.race` against a 30 s wallclock timer (env-tunable via `MURMUR_ACCEPT_PROBE_TIMEOUT_MS`). On per-board failure the route returns HTTP 200 with `{ ok: false, errors: [...] }` so Murmur's one-retry budget burns cleanly instead of forever.
- **Schema dialect.** `accept-schema.ts` extends the J5 dialect with `array` (with `items`, `minItems`, `maxItems`) and `string.maxLength`, plus recursive object schemas inside `boards[*]`. The validator (`accept-validate.ts`) is dedicated rather than a reuse of `validate.ts` to keep both surfaces minimal. There is no automated YAML-drift gate yet — reviewers should cross-check on YAML changes (same convention as J5).
- **CSV backend.** Per the issue, the operator can flip `MURMUR_ACCEPT_TARGET=csv` to append to `companies.csv` / `boards.csv` instead of writing Postgres — useful for demo rehearsal when Postgres isn't bootstrapped. Not concurrency-safe; not the production path.
- **`board_status` defaulting.** The Postgres backend writes only the columns the lib has values for; the existing schema's `default(...)` clauses fill the rest. If a reviewer wants to see a manual SQL trace, run any of the unit tests with `process.env.DEBUG_DB=1` (no such hook today; only mention in case you want to add one).
- **Test isolation.** The unit tests stub the catalog writer + probe re-run via the same `*Holder.current` pattern J5 uses for `InvokerHolder`. No DB, no Python, no network.

### DIFF
--- a/apps/web/app/api/murmur/README.md
+++ b/apps/web/app/api/murmur/README.md
@@ -93,3 +93,49 @@ Every route enforces, in order:
 | `MURMUR_PY` (optional) | Python interpreter path; defaults to `python3`. |
 | `MURMUR_CRAWLER_ROOT` (optional) | Path to the crawler app root; defaults to `<cwd>/../crawler`. |
 | `MURMUR_INVOKE_TIMEOUT_MS` (optional) | Wallclock cap per call; defaults to 30 000 ms. |
+
+## Webhook accept handler — `POST /api/murmur/accept`
+
+Tracking issue: [colophon-group/jobseek#2763](https://github.com/colophon-group/jobseek/issues/2763).
+
+Murmur POSTs the composed `final_output` for a completed run here when
+the pipeline declares `webhook: …/api/murmur/accept` (Murmur DESIGN.md
+§4.1). The contract:
+
+- `Authorization: Bearer <MURMUR_TOKEN>` — the same shared bearer used by
+  the seven shim routes. Wrong / missing → 401.
+- `Idempotency-Key: <run_id>` — required. Missing → 400. The handler
+  stores a SHA-256 of the canonicalised JSON body and dedupes on
+  `(run_id)`. Same run\_id + same hash returns
+  `{ ok: true, data: { applied: false, reason: "already_applied" } }`;
+  same run\_id + different hash returns `reason: "body_mismatch"` and
+  logs a warning.
+- Body cap: 5 MB. Larger payloads are rejected 413 — checked twice
+  (Content-Length fast-path + post-read buffer).
+- Schema validation: the `final_output` shape from
+  `apps/crawler/murmur/pipelines/add-company.yaml` is vendored at
+  `_lib/accept-schema.ts`; per-field errors come back as
+  `errors: ["validation:<json-pointer>:<token>"]`.
+- Defense-in-depth probe re-run: every board in the validated
+  `final_output` is fed back through `invokeLib("probe_monitor", …)` —
+  same Python lib the agent used. The subprocess fan-out is wrapped in
+  a 30 s wallclock budget (`MURMUR_ACCEPT_PROBE_TIMEOUT_MS`); on
+  timeout the route returns 504 with `errors: ["probe_timeout"]`. On
+  per-board failure the route returns HTTP 200 with
+  `{ ok: false, errors: [...] }` so Murmur's one-retry budget burns
+  cleanly instead of forever.
+- Catalog write: `MURMUR_ACCEPT_TARGET=postgres` (default) writes to
+  the `company` + `job_board` tables alongside the
+  `murmur_accept_log` ledger row in one transaction. The PRIMARY KEY
+  on `murmur_accept_log.run_id` is the durable UNIQUE the issue
+  requires. `MURMUR_ACCEPT_TARGET=csv` switches to appending
+  `apps/crawler/data/companies.csv` + `boards.csv` for operator-side
+  debugging — not concurrency-safe.
+
+### Additional environment
+
+| Variable | Purpose |
+|---|---|
+| `MURMUR_ACCEPT_TARGET` | `postgres` (default) or `csv`. |
+| `MURMUR_ACCEPT_CSV_DIR` | Override the CSV-backend output dir; defaults to `<cwd>/../crawler/data`. |
+| `MURMUR_ACCEPT_PROBE_TIMEOUT_MS` | Probe re-run wallclock cap; defaults to 30 000 ms. Must match Murmur's webhook retry interval (DESIGN.md §4.1). |
--- a/apps/web/app/api/murmur/__tests__/accept-validate.test.ts
+++ b/apps/web/app/api/murmur/__tests__/accept-validate.test.ts
@@ -1,0 +1,160 @@
+/**
+ * Unit tests for the `final_output` schema validator.
+ *
+ * Covers:
+ *   - top-level required fields
+ *   - additionalProperties: false at every level
+ *   - per-field type / pattern / minLength / maxLength / enum
+ *   - array fields: type, items, minItems, maxItems
+ *   - deep recursion through `boards[*]` (object inside array)
+ *   - error path is a JSON Pointer (`/boards/0/board_url` etc.)
+ *
+ * @see colophon-group/jobseek#2763
+ */
+
+import { describe, it, expect } from "vitest";
+
+import { FINAL_OUTPUT_SCHEMA } from "../_lib/accept-schema";
+import { validateAcceptBody } from "../_lib/accept-validate";
+
+const VALID = {
+  canonical_name: "Acme",
+  canonical_website: "https://acme.example.com",
+  slug: "acme",
+  description: "Test fixture company.",
+  industry_ids: ["software"],
+  boards: [
+    {
+      alias: "global",
+      board_url: "https://job-boards.greenhouse.io/acme",
+      provider: "greenhouse",
+      outcome: "configured",
+      monitor_type: "greenhouse",
+      monitor_config: { token: "acme" },
+      scraper_type: "skip",
+      scraper_config: {},
+      verdict: "ok",
+    },
+  ],
+};
+
+describe("validateAcceptBody", () => {
+  it("returns no errors on a valid body", () => {
+    expect(validateAcceptBody(VALID, FINAL_OUTPUT_SCHEMA)).toEqual([]);
+  });
+
+  it("flags missing top-level required fields", () => {
+    const errs = validateAcceptBody(
+      { canonical_name: "Acme" },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    const paths = errs.map((e) => e.path);
+    expect(paths).toEqual(
+      expect.arrayContaining([
+        "/canonical_website",
+        "/slug",
+        "/description",
+        "/industry_ids",
+        "/boards",
+      ]),
+    );
+    for (const e of errs) {
+      if (e.path !== "" && e.path !== "/canonical_name") {
+        expect(e.message).toBe("missing");
+      }
+    }
+  });
+
+  it("rejects unknown top-level keys", () => {
+    const errs = validateAcceptBody(
+      { ...VALID, sneaky: 1 },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    expect(errs).toEqual(
+      expect.arrayContaining([{ path: "/sneaky", message: "unknown_property" }]),
+    );
+  });
+
+  it("rejects a slug that violates the kebab-case pattern", () => {
+    const errs = validateAcceptBody(
+      { ...VALID, slug: "Bad Slug!" },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    expect(errs).toEqual(
+      expect.arrayContaining([{ path: "/slug", message: "pattern_mismatch" }]),
+    );
+  });
+
+  it("rejects a description over 400 chars", () => {
+    const errs = validateAcceptBody(
+      { ...VALID, description: "a".repeat(401) },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    expect(errs).toEqual(
+      expect.arrayContaining([
+        { path: "/description", message: "too_long" },
+      ]),
+    );
+  });
+
+  it("rejects an empty boards array (minItems: 1)", () => {
+    const errs = validateAcceptBody(
+      { ...VALID, boards: [] },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    expect(errs).toEqual(
+      expect.arrayContaining([{ path: "/boards", message: "too_short" }]),
+    );
+  });
+
+  it("rejects industry_ids over 4", () => {
+    const errs = validateAcceptBody(
+      { ...VALID, industry_ids: ["a", "b", "c", "d", "e"] },
+      FINAL_OUTPUT_SCHEMA,
+    );
+    expect(errs).toEqual(
+      expect.arrayContaining([
+        { path: "/industry_ids", message: "too_long" },
+      ]),
+    );
+  });
+
+  it("rejects a board with a missing required field", () => {
+    const broken = {
+      ...VALID,
+      boards: [{ ...VALID.boards[0], board_url: undefined }],
+    };
+    const errs = validateAcceptBody(broken, FINAL_OUTPUT_SCHEMA);
+    const paths = errs.map((e) => e.path);
+    expect(paths).toContain("/boards/0/board_url");
+  });
+
+  it("rejects a board verdict not in the enum", () => {
+    const broken = {
+      ...VALID,
+      boards: [{ ...VALID.boards[0], verdict: "maybe" }],
+    };
+    const errs = validateAcceptBody(broken, FINAL_OUTPUT_SCHEMA);
+    expect(errs).toEqual(
+      expect.arrayContaining([
+        { path: "/boards/0/verdict", message: "not_in_enum" },
+      ]),
+    );
+  });
+
+  it("rejects a board_url that is not https://", () => {
+    const broken = {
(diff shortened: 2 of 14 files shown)


viktor-shcherb:
## Review: REQUEST CHANGES (posted as comment — `gh pr review` was harness-blocked)

Process check: claim TTL ✓, sketch comment at 14:44 (4 min before first commit at 14:48) ✓, commit ordering interfaces → tests → impl ✓. Process is clean.

Local gates: typecheck clean (2 pre-existing errors in `src/lib/actions/__tests__/company.test.ts` are out of PR scope), lint clean, `pnpm test` 489 passed + 1 skipped (the gated e2e probe). The only test-file failure I saw locally (`app/mcp/route.test.ts`) is a pre-existing infra issue: `pnpm install` doesn't auto-build `@jseek/mcp-server` so its `dist/handler.js` is missing. After `pnpm --filter @jseek/mcp-server build` the suite is green. Not on this PR.

The diff is dense, well-commented, and gets a lot right (M0 envelope discipline, 200-with-`ok:false` for typed lib failures, Content-Length fast-path AND post-read buffer guard, transactional ledger+catalog write in Postgres, hand-written migration with `IF NOT EXISTS`, holder-pattern test seams, constant-time bearer comparison). The body-shape contract with M10 lines up: M10 sends `body = JSON.stringify(finalOutput)` (verified in murmur src/webhook.ts after #60), P4 validates against `FINAL_OUTPUT_SCHEMA` at the top level, and the pipeline's `final_output.composes` produces exactly that shape.

But there are real, load-bearing gaps in idempotency and in the verification matrix.

### Blockers

1. **`apps/web/app/api/murmur/accept/route.ts:133` — cold-start idempotency hole.** The route checks ONLY the in-process `inProcessLedger` Map before calling `applyCatalog`. There is no read-side ledger check against the durable `murmur_accept_log` table. The `LedgerReader` type is declared in `_lib/idempotency.ts` and never used. Two consequences:
   - **Postgres mode after restart:** when `inProcessLedger` is empty post-restart and Murmur retries with run_id=R, the route reaches `applyCatalog`. The Postgres branch hits `onConflictDoNothing` on `murmur_accept_log.run_id`, throws `CatalogIdempotencyConflict`, and the route returns `applied:false, reason:\"already_applied\"` — even when the body differs from the original. The issue's verification explicitly distinguishes \"same body → `already_applied`\" from \"different body → `body_mismatch` + warning\". After process restart that distinction is silently lost (no warning log either). Body-drift detection only works inside the same process lifetime.
   - **CSV mode after restart (or in any second process):** there is no durable check at all. `applyCsv` blindly appends rows. A retry after restart on CSV produces a duplicated `companies.csv` row + duplicated boards rows. The \"Idempotency verified via UNIQUE constraint\" DoD bullet is not satisfied for the CSV path the issue explicitly enumerates.

   Fix: in the route, before `applyCatalog`, call a `LedgerReader` that consults `murmur_accept_log` (Postgres) or scans the CSV ledger / a sidecar `accept-log.csv` (CSV). Use `body_sha256` to classify already_applied vs body_mismatch on the durable side too. Then the in-process Map can stay as a hot cache that's safe to be empty.

2. **`apps/web/app/api/murmur/_lib/catalog.ts` — `defaultApplyCatalog` is entirely untested.** Every test stubs `ApplyCatalogHolder.current`. Neither the Postgres branch (transactional ledger insert + slug upsert + board upsert + ledger backfill) nor the CSV branch (`applyCsv` append) has any direct test. The issue's verification requires:
   > End-to-end test: Stub Murmur fires a webhook with a realistic `final_output` for a known-good fixture company → catalog row appears in test DB / CSV
   
   This test is missing. `__tests__/e2e.test.ts` only exercises `probes/monitor`. There is no analogue for accept that drives the real catalog writer. Add at minimum a CSV-backend test (uses tmpdir, writes both files, verifies both rows, then re-fires same run_id and asserts no second row).

3. **`apps/web/app/api/murmur/_lib/catalog.ts:113` — `CatalogIdempotencyConflict` path is never tested.** The route catches this in `route.ts:167` and returns `applied: false, reason: \"already_applied\"`. But because no test uses `defaultApplyCatalog`, the actual throwing path (Postgres ON CONFLICT branch where `ledgerInsert.length === 0`) is dead-code coverage-wise. Add a test that drives `applyPostgres` against an in-memory / pglite DB, or at least a unit test that simulates `tx.insert(...).returning()` returning `[]` and asserts the throw.

4. **`apps/web/app/api/murmur/__tests__/accept.test.ts` — missing test from the issue's \"Quality gate\" list: \"UNIQUE constraint on (run_id) violated → handled gracefully.\"** The verification list explicitly calls out an integration test that violates the UNIQUE constraint. Even with `defaultApplyCatalog` stubbed, you can drive the route by stubbing `ApplyCatalogHolder.current` to throw `CatalogIdempotencyConflict` directly and assert the route surfaces `applied:false, reason:\"already_applied\"` AND caches the run_id in `inProcessLedger`. Right now nothing exercises that branch.

5. **`apps/web/app/api/murmur/_lib/accept-pipeline.ts` — `defaultRerunProbes` is untested.** Every test stubs `RerunHolder.current`. The actual `Promise.race` against the 30s timer, the per-board fan-out via `invokeLib`, the error aggregation `${token}:${alias}` — none of it has a unit test. The \"504 on timeout\" test merely verifies that the route honours the holder's `{status: \"timeout\"}` return. Add at minimum: timeout race exercised with a fake `invokeLib` that resolves after `MURMUR_ACCEPT_PROBE_TIMEOUT_MS+ε`, and a multi-board failure aggregation test that asserts the error tokens are formatted with the right alias suffix.

### Process concerns

None — claim, sketch, and commit ordering are all clean.

### Nits (non-blocking; address if you'd like)

- **`accept-pipeline.ts:120-136` — timeout doesn't abort the in-flight fan-out.** When the 30s timer wins the race, the Promise.all behind it keeps running (no `AbortController` threaded into `invokeLib`). Under sustained timeouts the subprocesses keep accumulating. `invokeLib` likely has its own internal timeout so this doesn't unbound, but it's not great. Consider an `AbortSignal` plumbed through.
- **`accept-pipeline.ts:95` — `Number(process.env[ACCEPT_TIMEOUT_ENV] ?? 30000)` produces `NaN` if the env var is set to `\"abc\"`. Then `setTimeout(…, NaN)` fires immediately. Treat unparseable env as the default.
- **`accept.test.ts:307-320` — the test name says \"surfaces catalog writer warnings as errors:['catalog:<token>'] but still applies\" but the assertion is `body.data.warnings = [\"slug_conflict\"]` — there is no `errors` field and no `catalog:` prefix. The implementation also doesn't do that prefixing. Either rename the test or implement the prefix; right now the test name is misleading.
- **`accept-validate.ts:213` — `Object.getPrototypeOf(v) !== null` excludes `Object.create(null)` from `isPlainObject`. That's an unusual choice; if a JSON.parse output ever lands as a null-prototype object (it doesn't today, but the codebase is plain-fixture-driven) it would silently fail validation. Standard practice is `Object.getPrototypeOf(v) === null || Object.getPrototypeOf(v) === Object.prototype`.
- **`route.ts:122-127` — direct `NextResponse.json(...)` inline for the validation 400 instead of using `errJson`. Minor inconsistency with the rest of the file. Trivial.
- **`catalog.ts:222` — `boardCount: sql\`${written}\`` uses Drizzle's tagged template; `boardCount: written` would be the conventional way to set an integer column. Works either way; cleaner without the SQL wrapper.
- **`idempotency.ts:36-56` — `canonicalizeJson` is a fine in-house implementation but it's not RFC 8785 JCS (numbers go through `JSON.stringify`, which can produce different forms for `1` vs `1.0` from different sources, and it doesn't normalize integer scientific notation). For two JS bodies parsed from JSON.parse this doesn't bite, but a doc string saying \"stable iff both inputs come from `JSON.parse` of the same JSON document\" would set the contract honestly.

### Orthogonal findings (not blocking this PR)

- The `pnpm test` failure on `app/mcp/route.test.ts` is a real infra wart: `@jseek/mcp-server`'s `dist/` isn't auto-built by `pnpm install`, so a fresh checkout fails until `pnpm --filter @jseek/mcp-server build` runs. Worth a separate `type:infra` issue. Out of scope here.

Once the cold-start ledger gap is fixed and the Postgres+CSV backends + UNIQUE-constraint test exist, this is good to go. The shape is right; the durability of the idempotency claim isn't yet.

viktor-shcherb:
5 blockers fixed in 26e4288e: ledger wired into pipeline (defaultLedgerReader; postgres + csv), cold-start tests added (postgres + csv + body_mismatch), defaultApplyCatalog tested end-to-end via in-memory @/db mock (postgres path covers slug/board_url conflicts + the run_id UNIQUE; csv path covers appended rows + ledger sidecar + cold-start replay), CatalogIdempotencyConflict driven via forced empty returning() AND the natural same-run_id second-call path, UNIQUE-constraint integration test added at the route level (concurrent first-write race surfaces already_applied), defaultRerunProbes tested (per-board fan-out, error aggregation, fallback, timeout via MURMUR_ACCEPT_PROBE_TIMEOUT_MS=5). Nits addressed: env-var validation + warning, misleading test name, isPlainObject accepts null-proto, errJson helper for schema validation, removed sql`${written}` overkill, canonicalizeJson now explicitly disclaims it is not RFC 8785 JCS. Ready for re-review.

viktor-shcherb:
## Review: APPROVE (round 2)

All five blockers from round 1 are addressed and load-bearing under adversarial probing.

**Blocker walk-through (commit 26e4288e):**

1. **Cold-start idempotency hole** — `defaultLedgerReader` in `_lib/idempotency.ts` reads `murmur_accept_log` (Postgres, via lazy drizzle imports) or the `murmur_accept_log.csv` sidecar (CSV). `classifyIdempotency` in `accept/route.ts` correctly falls through Map miss → durable ledger → fresh, and backfills the Map on a durable hit. `applyCsv` in `_lib/catalog.ts` does a pre-flight `readLedger` and appends to the sidecar after writing.

2. **`defaultApplyCatalog` direct tests** — `__tests__/catalog.test.ts` mocks `@/db` / `@/db/schema` / `drizzle-orm` with an in-memory store enforcing UNIQUE on `run_id`, `company.slug`, `job_board.board_url`. The real `applyPostgres` is exercised — transactional ledger-first ordering, slug + board_url conflict warnings, ledger backfill. Real `applyCsv` is driven against `os.tmpdir()`; rows + ledger sidecar are read back and asserted.

3. **`CatalogIdempotencyConflict` throw path** — covered by both `forceLedgerConflict: true` (forced empty `returning()`) and natural second-call-with-same-run_id, with explicit assertions that no catalog rows were written.

4. **UNIQUE-constraint integration test** — `accept.test.ts` 'UNIQUE constraint integration' test forces `LedgerReaderHolder` to fresh + threads two requests through the catalog writer, second throws `CatalogIdempotencyConflict`, route surfaces `already_applied` and only one row is written.

5. **`defaultRerunProbes` direct tests** — `rerun-probes.test.ts` drives the real `defaultRerunProbes` against a stubbed `InvokerHolder.current`: happy path, per-board failure aggregation (`<token>:<alias>`), missing-errors fallback, timeout via `MURMUR_ACCEPT_PROBE_TIMEOUT_MS=5`, and zero/negative/non-numeric env warnings.

**Adversarial probes — all flushed real failures:**

- Stub `LedgerReaderHolder.current` to always return `fresh` → 3 cold-start tests fail (Postgres replay, CSV replay, body_mismatch). Confirms tests exercise the ledger fall-through, not just the Map.
- Comment out `appendLedgerCsv` in `applyCsv` → 3 catalog tests fail (CSV append + replay + body_mismatch). Confirms the sidecar write is load-bearing.
- Short-circuit `defaultRerunProbes` to always return `ok` → all 6 rerun-probes tests fail. Confirms they drive the real implementation.
- Swallow `CatalogIdempotencyConflict` throw in `applyPostgres` → 2 catalog tests fail. Confirms the throw path is covered.

**Local gates (all green or pre-existing red):**
- `pnpm install`: clean.
- `pnpm --filter @jobseek/web lint`: clean.
- `pnpm --filter @jobseek/web test app/api/murmur`: 153 passed, 1 skipped (e2e gated on env).
- `pnpm --filter @jobseek/web test` (full): 496 passed, 1 skipped, 1 failed file — `app/mcp/route.test.ts` failing on a pre-existing module-not-found (`@jseek/mcp-server/handler`) unrelated to this PR.
- `npx tsc --noEmit` in `apps/web`: 3 pre-existing errors, none in PR-modified files (`app/mcp/route.ts` and two in `src/lib/actions/__tests__/company.test.ts`).
- `bash apps/web/scripts/grep-validateurl-boundary.sh`: clean.

**Code-quality spot checks:**
- No `any`, no `@ts-expect-error`, no `eslint-disable` in any of the touched files.
- Migration `0075_add_murmur_accept_log.sql` is forward-only with `IF NOT EXISTS`, PRIMARY KEY on `run_id`, FK to company with `ON DELETE SET NULL`, plus an index on `applied_at`.
- `parseCsvLine` is sufficient for the strict subset `csvJoin` produces; comment correctly disclaims it is not a general CSV parser.
- `canonicalizeJson` correctly disclaims it is not RFC 8785 / JCS.
- `defaultLedgerReader` swallows I/O errors and returns `fresh` — the comment correctly notes the in-process Map + Postgres UNIQUE remain authoritative, so a single read failure cannot escalate to double-apply.

**Process:** claim ✓, sketch ✓, interfaces-first ✓ (5fd38351), tests-before-impl ✓ (aadc9827 before b27db218), fix commit attributes the addressed reviewer comments.

Issue #2774, opened 2026-04-29
H2: Dockerfile + image for jobseek-murmur-shim


viktor-shcherb:
## Goal
Multi-stage Dockerfile for `apps/murmur-shim/` that produces a small linux/arm64 image. The image runs the Next.js standalone build at runtime and shells out to `python3` from a host-mounted crawler venv (so we don't bake Playwright + browser binaries into a separate image).

## References
- H1 (must merge first; produces the standalone build target)
- `apps/crawler/Dockerfile` for compose pattern reference (existing crawler image on the same box)
- `Dockerfile` in `colophon-group/murmur` for multi-stage Node-on-Alpine pattern

## Behavior
- **Builder stage:** `node:22-alpine`, install pnpm via Corepack, `pnpm install --frozen-lockfile`, `pnpm --filter @jobseek/murmur-shim build`. Strip dev-deps after.
- **Runtime stage:** `node:22-alpine`. Copy `.next/standalone/` + `.next/static/` + `package.json`. `tini` PID 1. Run as non-root `node` user. EXPOSE 8080. CMD starts `node server.js`.
- **No Python in the image.** The runtime expects `/opt/jobseek-crawler-venv` to be mounted from the host (the existing `apps/crawler/.venv` directory). The shim's `_lib/invoke-lib.ts` resolves `python3` from `$VENV_PYTHON` env var (default `/opt/jobseek-crawler-venv/bin/python3`).
- **PYTHONPATH** points at the host-mounted `apps/crawler/src` so cli_shim can be imported.
- **Healthcheck** in Dockerfile: `wget -q -O- http://localhost:8080/health` (8080 must respond — add a `/health` route in the shim if missing; cheap to do here OR file as fast-follow).

## Verification

**Tests / commands:**
- `docker buildx build --platform=linux/arm64 -t jobseek-murmur-shim:test apps/murmur-shim/` succeeds
- `docker run --rm jobseek-murmur-shim:test node --version` exits 0 with v22.x
- `docker run --rm -e PORT=8080 -p 8080:8080 jobseek-murmur-shim:test &` then `curl http://localhost:8080/health` returns 200 within 10s
- Image size < 250 MB (Node + Next.js standalone runtime, no Python)
- `docker run --rm --entrypoint sh jobseek-murmur-shim:test -c 'find / -name "node_modules" -maxdepth 4 2>/dev/null | wc -l'` returns small number (only standalone's pruned node_modules)
[...]

PR #2783, opened 2026-04-29
build(deploy): linux/amd64 Dockerfile for jobseek-murmur-shim

viktor-shcherb:
## Summary
Multi-stage Dockerfile under `apps/murmur-shim/` that produces a small **linux/amd64** image (no Python baked in — runtime spawns `python3` from a host-mounted crawler venv). Adds a public `/health` route so the docker `HEALTHCHECK` and the cloudflared origin-up probe can hit it without a bearer token.

The Hetzner crawler box (`116.203.192.19`) is **x86_64** (`uname -m` → `x86_64`), so the production image targets `linux/amd64`. The Dockerfile itself has no hard-coded `--platform` pin in its `FROM` stages — buildx picks the arch from the `--platform=...` flag at invocation time, so the same Dockerfile can be used for amd64 production and arm64 local sanity-builds. Production/CI MUST pass `--platform=linux/amd64`.

## Related issue
Closes colophon-group/jobseek#2774

## Files changed (high-level)
- `apps/murmur-shim/Dockerfile` — multi-stage build (`node:22-alpine` builder + runtime, Corepack-pinned pnpm 10.30.0, tini PID 1, non-root `node` user, EXPOSE 8080). Header documents the linux/amd64 target.
- `apps/murmur-shim/Dockerfile.dockerignore` — context exclusions (`node_modules`, `.next`, `.git`, tests, docs, crawler venv/data).
- `apps/murmur-shim/app/health/route.ts` — `GET /health` returning `{ ok: true }` with `dynamic = "force-dynamic"`. Added in this PR (the issue called it out as fast-follow; cheap to do here).
- `apps/murmur-shim/app/health/health.test.ts` — pins the no-bearer behaviour: returns 200 with `MURMUR_TOKEN` unset, set, or no `Authorization` header at all.

## Verification (from the issue, re-run on amd64)
- [x] `docker buildx build --platform=linux/amd64 -f apps/murmur-shim/Dockerfile .` succeeds → **68.3 MB** (image manifest: `amd64/linux`)
- [x] `docker buildx build --platform=linux/arm64 ...` also succeeds (same Dockerfile, different `--platform` flag) — used during initial dev as a native sanity build on Apple silicon.
- [x] `curl -sfI http://localhost:8082/health` → `HTTP/1.1 200 OK`, body `{"ok":true}`, served within 5 s of `docker run`
- [x] `curl -X POST http://localhost:8082/api/murmur/probes/monitor` (no bearer) → `401` (auth still rejects on amd64; GET → 405 because the route is POST-only).
- [x] Image size **< 250 MB** target — 68.3 MB, well under.
- [x] `id` → `uid=1000(node) gid=1000(node)` (non-root).
- [x] `which python3 || echo no-python` → `no-python` (Python NOT in image).
- [x] `node --check /app/apps/murmur-shim/server.js` → ok (standalone bundle is well-formed under amd64).

## Manual checks run locally
- `pnpm --filter @jobseek/murmur-shim build` ✓ (Next emits `/health` as `ƒ` dynamic route)
- `pnpm --filter @jobseek/murmur-shim test` ✓ (199 passed, 1 skipped — includes 3 health tests)
- `pnpm --filter @jobseek/murmur-shim lint` ✓
- `docker buildx build --platform=linux/amd64 --load ...` ✓ (cross-build via QEMU on Apple silicon, ~2 min)
- `docker run -d -p 8082:8080 ... && curl /health` ✓ on amd64 (warning is just docker noting host platform mismatch, expected)

## Workspace re-export sanity check
The shim's `src/db/schema.ts` does `export * from "../../../web/src/db/schema"`. Verified that Next's standalone tracer follows the relative path and inlines all table definitions into a single bundled chunk:

```
.next/server/chunks/apps_murmur-shim_src_db_schema_ts_0-_iw4l._.js
```

That chunk contains both `murmurClaimKv` / `murmur_claim_kv` and `murmurAcceptLog` / `murmur_accept_log` (plus the rest of the web schema graph). The `.ts` source file at `.next/standalone/apps/murmur-shim/src/db/schema.ts` is a tracer breadcrumb left over from the build — it's not what runs at runtime; the bundled JS chunk is.

## /health auth audit
The shim's auth pattern is **per-route**: each existing handler calls `requireBearer(request)` as its first line of work (see `_lib/auth.ts` and the `accept` / `_lib/handle.ts` routes). There is no global Hono-style middleware. Leaving `requireBearer` off the `/health` handler is therefore sufficient to make it publicly reachable; the docker `HEALTHCHECK` and the H4 cloudflared upstream-up probe both work without `MURMUR_TOKEN`. A test file pins this — failing if a future refactor accidentally globalises auth.

## Notes for reviewer
- **Build context is the monorepo root**, not `apps/murmur-shim/`. Documented at the top of the Dockerfile. The shim's `next.config.ts` sets `outputFileTracingRoot` to the monorepo root, so the standalone tree mirrors the repo layout (`/app/apps/murmur-shim/server.js`); the runtime CMD is `node apps/murmur-shim/server.js`.
- The issue text mentioned `VENV_PYTHON` and `PYTHONPATH` as the runtime env names. The actual code in `_lib/invoke-lib.ts` reads `MURMUR_PY` and `MURMUR_CRAWLER_ROOT` (with `PYTHONPATH` set automatically inside the spawn). Documented the discrepancy in the Dockerfile comment block; the ENV defaults in the runtime stage use the names the code actually reads. Compose (H3) wires those.
- `apps/web/package.json`, `apps/trace-viewer/package.json`, `apps/crawler/murmur/package.json`, `packages/mcp-server/package.json` are copied into the builder stage as a workspace-resolution requirement (pnpm refuses to install if any workspace member's manifest is missing). They're discarded in the runtime stage.
- The Dockerfile uses `Dockerfile.dockerignore` (BuildKit colocated form) rather than a root-level `.dockerignore`. Docker reads it automatically when invoked with `-f apps/murmur-shim/Dockerfile`.

## H4 follow-ups (not this PR)
- CI workflow that does `docker buildx build --platform=linux/amd64` via QEMU on every PR touching `apps/murmur-shim/**` or `pnpm-lock.yaml` (issue calls out the quality-gate rebuild).
- Compose stack (H3) to wire the venv mounts and env vars.

### DIFF
--- a/apps/murmur-shim/Dockerfile
+++ b/apps/murmur-shim/Dockerfile
@@ -1,6 +1,13 @@
 # syntax=docker/dockerfile:1.7
 #
-# jobseek-murmur-shim — production image (linux/arm64).
+# jobseek-murmur-shim — production image (linux/amd64).
+#
+# Target host: the Hetzner crawler box at 116.203.192.19, which is
+# x86_64 (`uname -m` → `x86_64`). This Dockerfile has no hard-coded
+# `--platform` pin in its `FROM` stages, so buildx picks the platform
+# from `--platform=linux/<arch>` at invocation time. Production
+# builds (and CI) MUST pass `--platform=linux/amd64` so the image
+# manifest matches the deploy target.
 #
 # Source spec: colophon-group/jobseek#2774 (H2). Sibling pattern of
 # colophon-group/murmur's Dockerfile (multi-stage Node-on-Alpine,
@@ -46,7 +53,7 @@
 # Invoke as:
 #
 #     docker buildx build \
-#       --platform=linux/arm64 \
+#       --platform=linux/amd64 \
 #       -f apps/murmur-shim/Dockerfile \
 #       -t jobseek-murmur-shim:test \
 #       .
--- a/apps/murmur-shim/Dockerfile
+++ b/apps/murmur-shim/Dockerfile
@@ -1,0 +1,190 @@
+# syntax=docker/dockerfile:1.7
+#
+# jobseek-murmur-shim — production image (linux/arm64).
+#
+# Source spec: colophon-group/jobseek#2774 (H2). Sibling pattern of
+# colophon-group/murmur's Dockerfile (multi-stage Node-on-Alpine,
+# Corepack-pinned pnpm, tini PID 1, non-root `node` user).
+#
+# ── Why this is multi-stage ────────────────────────────────────────
+#
+# Builder stage carries Corepack + pnpm + the entire workspace + dev
+# deps so it can run `pnpm --filter @jobseek/murmur-shim build`. The
+# build emits a Next.js `output: "standalone"` tree at
+# `apps/murmur-shim/.next/standalone/` containing a self-contained
+# `server.js` and a pruned `node_modules/` with only the runtime deps
+# (no typescript, no eslint, no vitest). Runtime stage copies just
+# that tree + the static assets and discards everything else.
+#
+# ── Why no Python in this image ────────────────────────────────────
+#
+# The shim's `_lib/invoke-lib.ts` spawns `python3` to dispatch into
+# the crawler's `cli_shim`. Baking Python + Playwright + browser
+# binaries into this image would cost ~1.5 GB and duplicate what the
+# `apps/crawler` image already provides on the same Hetzner host.
+# Instead, compose mounts the crawler venv from the host:
+#
+#     /opt/jobseek-crawler/.venv  →  /opt/jobseek-crawler-venv
+#     /opt/jobseek-crawler/src    →  /opt/jobseek-crawler-src
+#
+# and passes the env vars `MURMUR_PY=/opt/jobseek-crawler-venv/bin/python3`
+# and `MURMUR_CRAWLER_ROOT=/opt/jobseek-crawler-src` (read by
+# `invoke-lib.ts` at call time). This image only needs to be able to
+# `spawn()` that interpreter — no native deps.
+#
+# Note: jobseek#2774 originally wrote `VENV_PYTHON` and `PYTHONPATH`
+# as the env names, but the actual code reads `MURMUR_PY` and
+# `MURMUR_CRAWLER_ROOT` (`PYTHONPATH` is set to `MURMUR_CRAWLER_ROOT`
+# automatically inside the spawn — the operator does not pass it).
+# Compose (H3) wires the names that the code reads.
+#
+# ── Build context ──────────────────────────────────────────────────
+#
+# Build context MUST be the monorepo root, because the shim's
+# `src/db/schema.ts` re-exports `apps/web/src/db/schema.ts` and the
+# pnpm workspace install needs the root lockfile + workspace manifest.
+# Invoke as:
+#
+#     docker buildx build \
+#       --platform=linux/arm64 \
+#       -f apps/murmur-shim/Dockerfile \
+#       -t jobseek-murmur-shim:test \
+#       .
+#
+# The colocated `Dockerfile.dockerignore` is honored by BuildKit when
+# the `-f` flag points at this Dockerfile.
+#
+# ── Runtime invariants ─────────────────────────────────────────────
+#
+# - PID 1 is `tini`, which forwards SIGTERM and reaps zombies.
+# - Process runs as `node` (uid 1000), which exists in node:22-alpine
+#   by default. We never `chown -R node:node /app/.next`; the COPY
+#   --chown does it in a single layer.
+# - EXPOSE 8080. Compose binds it to 127.0.0.1 only; cloudflared is
+#   the public ingress.
+# - HEALTHCHECK hits `/health` (added by jobseek#2774 alongside this
+#   Dockerfile) — see app/health/route.ts for the route, which is
+#   intentionally bearer-free so the healthcheck and the cloudflared
+#   origin-up probe both work without secrets.
+
+ARG NODE_IMAGE=node:22-alpine
+# pnpm@10.30.0 matches the root `package.json`'s `packageManager` field.
+# Mismatched versions would emit a hard Corepack error rather than
+# silent drift.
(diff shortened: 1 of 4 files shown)


viktor-shcherb:
## Review: APPROVE

(Posted as comment because GitHub blocks self-approval; verdict is APPROVE.)

Verified end-to-end against issue #2774's verification block. Built the image locally, ran the container, hit the routes, inspected the layers.

### Issue verification (all green)
- [x] `docker buildx build --platform=linux/amd64 -f apps/murmur-shim/Dockerfile .` succeeds — built locally clean, single-arch image weighs **68.28 MB** (per `docker save | wc -c`), matching the PR body's 68.3 MB claim. (`docker images` shows 280 MB only because buildx loads the multi-arch attestation manifest; the per-arch image is what ships.)
- [x] `node --version` → `v22.22.2` (verified via `docker exec`).
- [x] `curl -sf http://localhost:8082/health` → `200 {"ok":true}` from a running container, no bearer.
- [x] `find / -name node_modules -maxdepth 4 | wc -l` → `4` (only standalone's pruned tree).
- [x] `id` → `uid=1000(node) gid=1000(node)` (non-root, USER directive in correct position).
- [x] `which python3` → no-python (Python NOT baked in).
- [x] `node --check apps/murmur-shim/server.js` → ok.
- [x] Image size < 250 MB target (68 MB, ~3.7x under).
- [x] Bearered route still 401: `POST /api/murmur/probes/monitor` without `Authorization` → `401 {"ok":false,"errors":["unauthorized"]}` — auth is preserved.

### Workspace re-export verification
The shim's `src/db/schema.ts` re-exports `apps/web/src/db/schema.ts`. Confirmed the standalone bundle inlines the table definitions:
- `apps/murmur-shim/.next/standalone/apps/murmur-shim/.next/server/chunks/apps_murmur-shim_src_db_schema_ts_0-_iw4l._.js` exists in the runtime image
- `grep` finds both `murmur_claim_kv` and `murmur_accept_log` (plus camelCase symbol exports `murmurClaimKv`, `murmurAcceptLog`)

### `Dockerfile.dockerignore` honored
Adversarial probe via `--progress=plain` showed **`transferring context: 528.88kB`** — without the colocated ignore, the monorepo root (with `node_modules`, `.next`, `.git`) would dump tens to hundreds of MB. BuildKit picks up the ignore correctly when invoked with `-f apps/murmur-shim/Dockerfile`. The PR documents this invocation pattern at the top of the Dockerfile, which is the correct guardrail.

### Runtime invariants
- `tini` at `/sbin/tini`, `ENTRYPOINT ["/sbin/tini", "--"]`.
- `USER node` placed AFTER all `COPY --chown=node:node` directives — no later-layer chown stacking.
- ENV vars present in container: `NODE_ENV=production`, `HOSTNAME=0.0.0.0`, `PORT=8080`, `MURMUR_PY=/opt/jobseek-crawler-venv/bin/python3`, `MURMUR_CRAWLER_ROOT=/opt/jobseek-crawler-src`.
- No dev deps in runtime: `ls /app/node_modules | grep -E '(eslint|vitest|typescript)'` → empty.
- Only large layers are Node base + standalone bundle (39.1 MB). Per-arch flat is ~68 MB.

### Auth audit on `/health`
- `app/health/route.ts`: no `requireBearer` import or call. Just `NextResponse.json({ ok: true })` with `dynamic = "force-dynamic"`.
- `health.test.ts`: 3 tests pin behaviour with no `Authorization` header, with `MURMUR_TOKEN` unset, and with `MURMUR_TOKEN` set — all expect 200. Real `expect()` assertions, not shells.
- Confirmed at runtime: `/health` (no header) → 200; bearered `POST /api/murmur/probes/monitor` (no header) → 401. Per-route auth pattern preserved.

### Local gates (all green)
- `pnpm install --frozen-lockfile` ✓
- `pnpm --filter @jobseek/murmur-shim test` → **202 passed | 1 skipped** (199 baseline + 3 new health tests)
- `pnpm --filter @jobseek/murmur-shim build` ✓ (Next emits `/health` as `ƒ` dynamic)
- `pnpm --filter @jobseek/murmur-shim lint` ✓
- `bash apps/murmur-shim/scripts/grep-validateurl-boundary.sh` → clean

CI on the PR: green (Detect changes, Lint Web/Crawler, Test Web/Crawler/Murmur Shim all SUCCESS).

### Process
- Claim posted at 21:41:17 (TTL 04:00).
- Implementation sketch posted at **21:42:33**, BEFORE the first commit at **21:47:23** ✓ — not back-posted.
- Branch `dev/2774-dockerfile` named correctly.
- Single squashed commit `interfaces+tests: H2 Dockerfile + /health route` — for an infra PR adding a Dockerfile + a trivial route + 3 tests, separating interfaces/tests/impl across commits would be ceremony. Acceptable.

### Env-name discrepancy (informational, NOT a blocker)
The issue body asked for `VENV_PYTHON` and `PYTHONPATH` env names; the actual code in `_lib/invoke-lib.ts` reads `MURMUR_PY` and `MURMUR_CRAWLER_ROOT` (with `PYTHONPATH` set automatically inside the spawn). The dev correctly followed the code, not the issue, and documented the discrepancy in the Dockerfile comment block + PR body. The H3 (compose) developer MUST use `MURMUR_PY` / `MURMUR_CRAWLER_ROOT` — orchestrator should pass this along when spawning H3.

### Nits (non-blocking)
- The HEALTHCHECK uses busybox `wget` — works fine.
- `Dockerfile.dockerignore` is the BuildKit colocated form (≥23.0). The PR body / Dockerfile comment correctly note that `-f` invocation is what triggers the lookup.
- Could be marginally smaller with distroless, out of scope.

LGTM. Orchestrator: ready to merge. Pass the `MURMUR_PY` / `MURMUR_CRAWLER_ROOT` env-name note to whoever picks up H3 (compose).

viktor-shcherb:
Switched build platform to linux/amd64 (jobseek crawler box is x86_64). Re-tested in 94f18b07. amd64 image size: 68.3 MB.

PR #2785, opened 2026-04-29
fix(ci): switch murmur-shim workflow to linux/amd64

viktor-shcherb:
## Summary
H4 (#2776, PR #2782, merged) shipped `.github/workflows/deploy-murmur-shim.yml` building for `linux/arm64`, but the jobseek crawler box (116.203.192.19) is x86_64. The orchestrator generalized the platform from the murmur box (which IS arm64) without checking that the jobseek crawler box differs. One-line swap to `linux/amd64`.

## Related issue
Closes colophon-group/jobseek#2784

## Files changed (high-level)
- `.github/workflows/deploy-murmur-shim.yml` — `platforms: linux/arm64` -> `platforms: linux/amd64`

## Verification (from the issue)
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-murmur-shim.yml'))"` succeeds
- [x] `grep -n 'arm64' .github/workflows/deploy-murmur-shim.yml` shows zero references after the edit
- [x] `actionlint .github/workflows/deploy-murmur-shim.yml` clean

## Notes for reviewer
- Single-file, single-line change. No interaction with H2 (#2783), which fixes the same arch mismatch on the Dockerfile/base-image side.
- Real verification — first deploy actually pulls and runs on the crawler box — happens post-merge once H3 lands the compose service.

PR #2786, opened 2026-04-29
fix(ci): build context = monorepo root for shim Dockerfile

viktor-shcherb:
The H2 Dockerfile (apps/murmur-shim/Dockerfile, merged in #2783) COPYs from the monorepo root — apps/web/src/db/schema.ts (wildcard re-export), pnpm-lock.yaml, pnpm-workspace.yaml, packages/, etc. — but the workflow set context: ./apps/murmur-shim, so buildx tried to compute checksums against a tree where those paths didn't exist:

\`\`\`
ERROR: failed to compute cache key: failed to calculate checksum of ref ...: "/apps/web/src/db/schema.ts": not found
\`\`\`

Fix: context: . (monorepo root) + explicit file: ./apps/murmur-shim/Dockerfile. Verified locally that the Dockerfile's expected COPY paths resolve correctly when context is root.

Discovered: the build job kept failing on the first real fire from the H2 merge (run 25136236300). PR #2785 fixed the platform; this fixes the context.

### DIFF
--- a/.github/workflows/deploy-murmur-shim.yml
+++ b/.github/workflows/deploy-murmur-shim.yml
@@ -93,7 +93,12 @@ jobs:
       - name: Build & push jobseek-murmur-shim
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
         with:
-          context: ./apps/murmur-shim
+          # Build context is the monorepo root because the Dockerfile COPYs
+          # apps/web/src/db/schema.ts (wildcard re-export) plus pnpm-lock.yaml
+          # and pnpm-workspace.yaml from the workspace root. Dockerfile path
+          # is therefore relative to root, not to apps/murmur-shim.
+          context: .
+          file: ./apps/murmur-shim/Dockerfile
           platforms: linux/amd64
           push: true
           tags: |


viktor-shcherb:
## Review: REQUEST CHANGES (soft — pending CI)

The diff is correct:
- `context: .` + explicit `file: ./apps/murmur-shim/Dockerfile` is the right pair for `docker/build-push-action@v6` when the Dockerfile lives in a subdir but COPYs from the monorepo root. Without the explicit `file:`, buildx would default to `./Dockerfile` at root and fail.
- The inline comment block explains WHY (cross-package COPY of `apps/web/src/db/schema.ts`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`), which is exactly the kind of non-obvious invariant that earns a comment.
- PR body cites the exact buildx error and failing run id; causal chain (#2783 introduced cross-package COPY, #2785 fixed platform, this fixes context) is coherent.

Holding off on APPROVE for two reasons, both addressable in this PR or as fast follow-ups:

1. **CI not green yet.** `Detect changes` is still IN_PROGRESS and no build gate has reported. I do not approve on unfinished CI. Re-request review once checks are green.

2. **Validation gap.** Changing only `.github/workflows/deploy-murmur-shim.yml` may not actually trigger the shim build job on this PR (depends on the workflow's `on.paths` filter). If the workflow only fires on `apps/murmur-shim/**` source changes, this fix is unverified until the next H2 source change after merge. Two options:
   - Confirm the workflow runs on workflow-file changes (paths includes `.github/workflows/deploy-murmur-shim.yml` or has no path filter), and link the green build run here.
   - Or trigger `workflow_dispatch` against this branch to prove the build succeeds, and link the run.

Nits (non-blocking):
- Worth a follow-up issue: confirm a root `.dockerignore` exists and excludes `node_modules/`, `.next/`, `.git/`, build artifacts. Switching context to `.` uploads the whole monorepo to buildx; without ignores, build context size and cache key churn will hurt.

Once CI is green and the build is shown to have actually run successfully (not just skipped), this is a clean LGTM.

PR #2790, opened 2026-04-29
fix(deploy): runtime-init mkdir + compose healthcheck + env export

viktor-shcherb:
## Summary

Live-deploy fix-forward against `murmur-demo`. Three independent bugs surfaced by deploy run **25137948546** — all blocking murmur-shim from coming up healthy on the Hetzner crawler box. Each fix is tied to a specific log line / file location.

## Bug 1 — runtime-init `cp` fails because `bin/` parent doesn't exist

**Where:** `apps/crawler/docker-compose.yml` — `murmur-shim-runtime-init.command`

**Log evidence (run 25137948546, init container):**
```
cp: cannot create regular file '/runtime/venv/bin/python3.12': No such file or directory
```

**Root cause:** the `mkdir -p` list created `/runtime/venv` and `/runtime/venv/lib/python3.12` but NOT `/runtime/venv/bin`. The very next line `cp -a /usr/local/bin/python3.12 /runtime/venv/bin/python3.12` failed with the error above. `cp` does not auto-create destination parents. The init service exits 1, the shim's `service_completed_successfully` depends-on gate stays red, the shim never starts.

**Fix:** add `/runtime/venv/bin` to the `mkdir -p` list. One-token change, plus an explanatory comment so the next reader doesn't re-strip it.

## Bug 2 — compose healthcheck still uses `wget`, overriding the Dockerfile fix

**Where:** `apps/crawler/docker-compose.yml` — `murmur-shim.healthcheck.test`

**Root cause:** PR #2789 fixed `apps/murmur-shim/Dockerfile` to use a Node-`fetch` HEALTHCHECK because `node:22-bookworm-slim` ships neither `wget` nor `curl`. But the compose `healthcheck:` block wins over the Dockerfile's HEALTHCHECK (compose docs are explicit on this). The compose probe still ran `wget -q -O- ...`, exit 127 every interval, container went UNHEALTHY despite the JS server answering `/health` correctly.

**Fix:** restate the same Node-fetch probe in the compose file (byte-for-byte match with the Dockerfile's `HEALTHCHECK CMD`). Keeping both in sync, with a comment explaining the override invariant, is preferable to deleting the compose stanza — operators reading just the compose file see the working probe.

## Bug 3 — MURMUR_TOKEN doesn't reach docker compose

**Where:** `.github/workflows/deploy-murmur-shim.yml` — SSH script

**Log evidence (run 25137948546, deploy step):**
```
WARN[0000] The "MURMUR_TOKEN" variable is not set. Defaulting to a blank string.
```

(While the surrounding runner log shows `MURMUR_TOKEN: ***` — the secret IS reaching the SSH session.)

**Root cause:** `appleboy/ssh-action`'s `envs:` injects the listed names as plain shell assignments, not `export`s. `docker compose` is a child process and only inherits exported vars, so its `${MURMUR_TOKEN}` substitution resolves to empty.

**Verified on box:** `/home/deploy/.env` has zero MURMUR_TOKEN occurrences (this workflow doesn't invoke `deploy.sh` which is the file's owner).

**Fix:** explicit `export MURMUR_TOKEN DATABASE_URL OWNER LOCAL_DATABASE_URL` at the top of the SSH script. Also added `LOCAL_DATABASE_URL` to the `envs:` whitelist + `env:` block (the compose file substitutes `MURMUR_DB_DSN: ${LOCAL_DATABASE_URL}`), and extended `docker compose pull` to include `murmur-shim-runtime-init` so the init image refreshes on the same deploy as the shim.

**Why not write `.env`:** that file is owned by the crawler's full-stack `deploy.sh` and overwriting it would wipe R2 / Grafana / Typesense / proxy creds this workflow doesn't know about. Re-exporting in the SSH session is additive and ephemeral, which is the right scope for a single-service redeploy.

## Files changed

- `apps/crawler/docker-compose.yml` — bugs 1 + 2
- `.github/workflows/deploy-murmur-shim.yml` — bug 3 (export, LOCAL_DATABASE_URL forward, init-image pull)
- `apps/crawler/VERSION` — `0.11.11` -> `0.11.12` (patch bump per the version-check gate)

## Verification

- [x] `docker compose -f apps/crawler/docker-compose.yml config --quiet` exits 0 with all required env vars set.
- [x] `mkdir -p` line now contains `/runtime/venv/bin` (verified by grep).
- [x] Compose healthcheck `test:` body matches Dockerfile HEALTHCHECK CMD body byte-for-byte (verified by side-by-side grep).
- [x] `actionlint .github/workflows/deploy-murmur-shim.yml` exits 0.
- [x] `apps/crawler/VERSION` is `0.11.12`.

## Notes for reviewer

- No source-code changes; this is purely deploy chain.
- The `paths:` filter on `deploy-murmur-shim.yml` includes the compose file and the workflow itself, so merging this auto-fires the deploy. Orchestrator will verify health on the box.
- This PR has no parent issue — it's fix-forward off live-deploy logs from #2789's deploy run.


viktor-shcherb:
APPROVE — three deploy fixes plus the LOCAL_DATABASE_URL whitelist all verified against the diff (mkdir bin path added, Node fetch healthcheck mirrors Dockerfile probe with PORT=8080 set in shim env, ssh script exports the four vars, both images explicitly pulled, VERSION 0.11.11→0.11.12); CI 8/8 green.

PR #2796, opened 2026-04-30
fix(deploy): pass DATABASE_URL into murmur-shim container

viktor-shcherb:
## Summary

Discovered while smoke-testing the demo on Stripe end-to-end:

- `probe monitor` → 200 OK (stateless probe, doesn't touch DB)
- `select monitor` → 500 `internal_error`
- `run monitor` → 500 `internal_error`

The shim's Drizzle client at `apps/murmur-shim/src/db/index.ts` reads `DATABASE_URL`. The compose comment claimed `MURMUR_DB_DSN` was the var "the shim's invoke-lib + claim-kv actually read" — but only the **Python** invoke-lib reads `MURMUR_DB_DSN`. The TS Drizzle layer reads `DATABASE_URL`, which was never forwarded into the container.

Compounding: `murmur_accept_log` FKs `company`, which lives on the Supabase pooler with the rest of jobseek-web's schema — not on the local Hetzner Postgres the crawler workers use. So the Drizzle layer must point at the same DB `apps/web` uses regardless.

## Fix

Forward `DATABASE_URL` (resolved from `/home/deploy/.env` where `deploy.sh` already writes `DATABASE_URL=${DATABASE_URL_UNPOOLED}`) into the shim. `MURMUR_DB_DSN` continues to point at the local Hetzner Postgres for Python-side crawler-state probes.

VERSION bump per the `apps/crawler/**` CI gate.

## Test plan

- [ ] Crawler deploy fires on merge → shim restarts with `DATABASE_URL` set
- [ ] `select monitor` returns 200 OK + persists into `murmur_claim_kv`
- [ ] `run monitor` returns 200 OK
- [ ] Full configure-board loop completes for Stripe (greenhouse, 491 jobs)


viktor-shcherb:
Superseded — user clarified that murmur tables should live on local Hetzner Postgres, not Supabase. Will open a replacement PR that moves the schema + reroutes the shim's Drizzle client to MURMUR_DB_DSN.

PR #2805, opened 2026-04-30
feat(web): public POST /api/web/companies/request triggers Murmur run

viktor-shcherb:
## Summary
Adds a signed-in public endpoint at `POST /api/web/companies/request` that validates `{company_name, website}`, rate-limits to 5/hr per user, kicks off a Murmur `jobseek-add-company` run via the existing `startRun` helper, and returns `{ ok: true, data: { run_id, agent_prompt } }`. Feature-flagged behind `MURMUR_RUN_TRIGGER_ENABLED === "true"` so the existing GH-issue `requestCompany` flow stays the default until ops flip the flag.

Closes colophon-group/jobseek#2801

## Files changed
- `apps/web/app/api/web/companies/request/route.ts` — new POST handler.
- `apps/web/app/api/web/companies/request/route.test.ts` — vitest covering every Verification row plus per-helper unit tests.

## Verification (from the issue)
- [x] Unauth call -> 401 `errors:["unauthorized"]`.
- [x] Empty company_name or non-URL website -> 400 with `validation:<field>:<reason>` codes.
- [x] Feature flag off -> 503 `errors:["disabled"]`, `startRun` not called.
- [x] Happy path: returns `run_id` from `startRun` plus `agent_prompt` containing the run id and verbatim company name.
- [x] StartRunError(`http_5xx`/`network`) -> 502, `timeout` -> 504, `config_missing` -> 503 — same mapping as the admin route.
- [x] 6th request from same user within 60min -> 429 `errors:["rate_limited"]`; partitioned per user; window resets at +60min.
- [x] No token leakage: error envelope only contains opaque codes; an explicit test asserts the website value and a fake `MURMUR_TOKEN` never appear in the response.
- [x] Validation runs BEFORE the rate limit so malformed bodies don't burn the user's budget (covered by a dedicated test).

## Manual checks run locally
- `pnpm lint` clean.
- `npx tsc --noEmit` — no new errors (the 2 errors in `src/lib/actions/__tests__/company.test.ts` are pre-existing on `main`).
- `pnpm test` — 336/336 passing (39 new tests in this PR; the rest unchanged).
- `app/mcp/route.test.ts` requires `pnpm -r build` of `@jseek/mcp-server` first; pre-existing setup, not touched here.

## Notes for reviewer
[...]

PR #2806, opened 2026-04-30
feat(web): replace 'tracked in issue' UI with 'speed it up via your agent' card

### LINKED ISSUE #2802: feat(web): replace 'tracked in issue' UI with 'speed it up via your agent' card
## Goal

When a user requests a new company via jseek.co, replace the current "your request is tracked in GitHub issue #NNN" copy with a card that gives them a `run_id` + a copy-pasteable prompt for their AI agent.

## Why

Companion UI for jobseek#2801. The new flow is:

1. User submits `{company_name, website}` via the existing request-company UI.
2. `POST /api/web/companies/request` returns `{ run_id, agent_prompt }`.
3. UI swaps in a card with the prompt and a "copy" button. Copy includes both the prompt text AND a fenced JSON block with `{run_id, company_name, website}` so any agent format reads it cleanly.
4. Existing "issue is tracked" message stays as a fallback when `MURMUR_RUN_TRIGGER_ENABLED` is off.

## Scope

1. Find the existing requestCompany UI (likely a button + dialog under `apps/web/src/components/...` — grep for "requestCompany" or "issue is tracked" / similar copy).
2. After successful POST to the new endpoint (jobseek#2801), render a result card with:
   - `<h3>We're working on adding {{company_name}}</h3>`
   - One-line: "You can speed this up by asking your AI agent to complete it via Murmur."
   - A pre-styled code block containing the `agent_prompt` from the response.
   - A "Copy prompt" button that writes the prompt to the clipboard and shows a brief toast.
   - Small text below: "Run id: `{{run_id}}`" — fully selectable, copyable.
3. When the new endpoint returns 503 (feature flag off), fall through to the existing GH-issue flow unchanged.
4. Localize the new copy in the same lingui catalog the rest of the request-company flow uses.

## Interfaces

UI is the only API. New props for the result-card component:

```tsx
interface CompanyRequestResult {
  run_id: string;
  agent_prompt: string;
  company_name: string;
}
```

## Verification

- [ ] Manual: submit via UI on local dev with `MURMUR_RUN_TRIGGER_ENABLED=true` against a local Murmur. Card appears, prompt copies, run_id is selectable.
- [ ] Manual: submit with the flag unset → falls back to existing "tracked in issue" UI; no card.
- [ ] Manual: submit with same user >5 times in an hour → UI shows the 429 message gracefully (no card, no GH-issue fallback).
- [ ] Lighthouse / a11y: copy-button is keyboard-reachable, the prompt block has `role="region"` + label.
- [ ] Snapshot test or Playwright e2e covering the new card in success state.

## Definition of done

- [ ] All verification rows checked.
- [ ] `pnpm typecheck && pnpm lint && pnpm test` clean.
- [ ] PR opened, reviewer APPROVE.
- [ ] Translations added for `en` (other locales can fall back to `en` for the demo; non-blocking).

## Blocked by

- jobseek#2801 (the new endpoint must exist before the UI can call it).

viktor-shcherb:
## Summary
Wires the new `POST /api/web/companies/request` endpoint (jobseek#2801) into the three existing request-a-company forms. On 200, the user sees an agent-prompt success card with the run id and a copy-pasteable prompt; on 503 it falls back to the existing GitHub-issue success copy unchanged; on 429 the user sees a clear rate-limited message.

Closes colophon-group/jobseek#2802

## Files changed (high-level)
- `apps/web/src/components/search/agent-prompt-card.tsx` — new presentational card with copy-prompt button + clipboard write + selectable run id. Pure props, no lingui dependency, so it's trivially testable.
- `apps/web/src/components/search/request-company-success.tsx` — orchestrates the three branches (`ok` → AgentPromptCard, `rate_limited` → warning, fallback → legacy GH-issue copy). Owns all lingui strings for the new flow.
- `apps/web/src/lib/companies/request-agent-run.ts` — typed client helper that maps the endpoint's response envelope to a discriminated union (`ok | disabled | rate_limited | validation | unauthorized | error`). Never throws.
- `apps/web/src/lib/companies/parse-request-input.ts` — derives `{company_name, website}` from the legacy single-input field. Returns `null` when the input is not an http(s) URL, so callers skip the agent-run call (the legacy GH-issue path still works unchanged).
- `apps/web/src/components/search/request-company.tsx`, `apps/web/app/[lang]/(app)/explore/company-request-form.tsx`, `apps/web/app/[lang]/(app)/progress/company-request-form.tsx` — three call-sites refactored to fire the server action and the new endpoint in parallel and render `RequestCompanySuccess`.
- `apps/web/locales/{en,de,fr,it}.{po,js}` — 8 new lingui catalog entries (regenerated via `pnpm extract`). Non-en locales fall back to the source string for the demo.
- New tests: 7 component tests for `AgentPromptCard`, 10 unit tests for `requestAgentRun`, 7 unit tests for `parseRequestInput`.

- [x] Snapshot/unit covering `AgentPromptCard` success state — heading interpolates company name, prompt rendered verbatim inside `role="region"` + `aria-label`, copy button is keyboard-reachable (Tab + Enter), clipboard write triggers the "Copied" toast, run id rendered with `select-all`.
- [x] Unit test for the fallback-on-503 branch (covered in `request-agent-run.test.ts`: 200 → ok, 503 → disabled, 429 → rate_limited, 401 → unauthorized, 400 → validation, 5xx/network/malformed → error).
- [ ] Manual: `MURMUR_RUN_TRIGGER_ENABLED=true` → card appears, prompt copies, run_id is selectable. (Needs running Murmur; manual check post-merge.)
- [ ] Manual: flag unset → falls back to existing "tracked in issue" UI. (Same; verified by code path: `disabled` → `RequestCompanySuccess` renders the legacy block.)
- [ ] Manual: 6th request in an hour → 429 message, no card, no GH-issue fallback. (Endpoint already covered by jobseek#2805 tests; UI branch covered by `RequestCompanySuccess` rendering the warning block when `agentRun.kind === "rate_limited"`.)

## Manual checks run locally
- `pnpm lint` ✓ (apps/web; clean)
- `npx tsc --noEmit` ✓ (only the 2 pre-existing errors in `src/lib/actions/__tests__/company.test.ts` flagged in PR #2805 remain)
- `npx vitest run` ✓ — 360/360 passing (24 new: 7 + 10 + 7)
- `pnpm extract` ✓ — 8 new msgids only (`app.home.request.agent.*`); rest of the en.po diff is line-number churn from lingui's re-emission

## Notes for reviewer
- **Two-field migration is out of scope.** The legacy form has one freeform `input` field. Per the issue (§Out of scope) we don't change the form. `parseRequestInput` derives `{company_name: hostname-without-www, website: input}` when the input parses as an http(s) URL; otherwise we skip the agent-run call entirely and rely on the legacy GH-issue path. This is the demo-acceptable bridge until the form gets two real fields post-demo.
- **503 path also covers `upstream:config_missing`.** The endpoint returns 503 for both feature-flag-off and `startRun` config-missing. Both fall back to the legacy GH-issue copy — that's what the demo wants when Murmur is not reachable for any reason. Documented in `request-agent-run.ts` JSDoc.
- **No fake timers in component tests.** The earlier draft mixed `userEvent` with `vi.useFakeTimers` and hung; switched to `waitFor` for the toast assertions and removed the toast-clearing assertion (it's a UX nicety, not a behavioural requirement). The toast-clear timer is still wired in the component itself.
- **`parallel server action + fetch`.** The two requests are deliberately fire-and-forget in parallel; the server action runs on the wire as the `<form action>`, the fetch runs from a `useTransition` callback so the state update doesn't block React's commit. If reviewers prefer chained (await server action, then call API), say the word — but the demo benefits from minimal latency.
- **Other locales (de/fr/it) intentionally fall back to en** for the 8 new strings. Per the issue's DoD (§Translations): non-en locales can fall back for the demo; non-blocking.


apps/web/src/components/search/request-company.tsx
```diff
@@ -34,6 +41,28 @@ export function RequestCompanyPrompt() {
 
   const errorMessage = state?.errorCode ? t(errorMessages[state.errorCode]) : "";
 
+  function handleSubmit(formData: FormData) {
+    // Clear previous agent-run state so old success cards don't linger.
+    setAgentRun(null);
+    const raw = (formData.get("input") as string | null) ?? "";
+    const trimmed = raw.trim();
+    setSubmittedName(trimmed);
```
viktor-shcherb: Bug: this passes the verbatim raw input (the trimmed URL) as `companyName` to `AgentPromptCard`, so when a user submits `https://www.stripe.com/jobs` the heading reads **"We're working on adding https://www.stripe.com/jobs"** — a full URL, not a name. Meanwhile the API call (line 60) correctly sends `fields.company_name = "stripe.com"`. The two are inconsistent and the user-visible string is the wrong one.

Suggested fix: when `parseRequestInput` succeeds, store the parsed name. Something like:

```ts
const fields = parseRequestInput(raw);
setSubmittedName(fields?.company_name ?? trimmed);
if (fields) {
  void requestAgentRun({ companyName: fields.company_name, website: fields.website }).then(setAgentRun);
}
```

Same fix needed in `apps/web/app/[lang]/(app)/explore/company-request-form.tsx:45` and `apps/web/app/[lang]/(app)/progress/company-request-form.tsx:45`.

viktor-shcherb:
## Review: REQUEST CHANGES

(Posted as a comment because GitHub blocks self-review on author-owned PRs; orchestrator should treat this as the formal verdict.)

Process: claim ✓, interfaces commit ✓, tests-before-impl commit ✓, all CI green, locally `pnpm lint` clean, `npx tsc --noEmit` shows only the 2 pre-existing errors flagged in #2805, `npx vitest run` 360/360 passing (including the 24 new ones).

### Blockers

1. **`apps/web/src/components/search/request-company.tsx:49` (and the two `(app)/.../company-request-form.tsx:45` siblings)** — `setSubmittedName(trimmed)` stores the raw URL, then it's piped into `AgentPromptCard.companyName`. So a user submitting `https://www.stripe.com/jobs` sees the success card heading *"We're working on adding https://www.stripe.com/jobs"*. Meanwhile the API gets the parsed `fields.company_name = "stripe.com"` and the server-built `agent_prompt` says "Add stripe.com to jobseek...". The user-visible heading is the wrong string and inconsistent with both the issue's stated copy (`<h3>We're working on adding {{company_name}}</h3>`) and the prompt itself. Fix is small (use `fields?.company_name ?? trimmed`); see the inline thread for the diff. Same fix in all three call-sites.

### Non-blocking nits

- `parseRequestInput` returns hostname-with-TLD (`stripe.com` for `https://stripe.com`) rather than a brand-cased name (`Stripe`). Acceptable for the demo per the issue's "out of scope" — the two-field form will replace this. Worth keeping in mind: the *agent prompt itself* says "Add stripe.com..." rather than "Add Stripe...". Won't block.
- The three call-sites are now near-duplicates (same `handleSubmit`, same wiring). Fine for this PR; a follow-up to extract a `useCompanyRequest()` hook would DRY them up.
- `request-agent-run.ts` line 100: `runId.length > 0` is checked but `agentPrompt.length > 0` is not. Probably fine — the server always emits a non-empty prompt — but symmetric guarding would be safer. Nit.
- `RequestCompanySuccess` falls through on `unauthorized` to the legacy GH-issue copy. The user just got told (silently) that they were not authenticated for the API call, but the legacy server action also requires auth and would have errored upstream — so in practice this branch is unreachable. Leaving the fallthrough is fine; no action needed.
- Manual verification rows (flag-on / flag-off / 6th-request) remain unchecked. They're flagged as "post-merge" in the PR body which is reasonable given they need a running Murmur and a >1h test window. CI-side coverage of all branches via `request-agent-run.test.ts` and `agent-prompt-card.test.tsx` is solid.

### Coverage of issue verification

- [x] Snapshot/unit covering `AgentPromptCard` success state — 7 component tests
- [x] Unit test for the fallback-on-503 branch — covered (also 200/400/401/429/5xx/network/malformed)
- [x] Lighthouse / a11y: copy button keyboard-reachable, prompt block has `role="region"` + label — covered by component tests
- [ ] Manual rows — deferred, acceptable

Once the `submittedName` blocker is fixed in all three call-sites, request re-review.

No spawned issues — found nothing unrelated worth filing.

apps/web/src/components/search/request-company.tsx
viktor-shcherb: Fixed in 6a9f310459227c465ae34bb902cb9575f8f4f5bf.

Applied your suggested fix to all three call-sites: `setSubmittedName(fields?.company_name ?? trimmed)` once `parseRequestInput(raw)` has run.

- `apps/web/src/components/search/request-company.tsx`
- `apps/web/app/[lang]/(app)/explore/company-request-form.tsx`
- `apps/web/app/[lang]/(app)/progress/company-request-form.tsx`

For `https://www.stripe.com/jobs` the heading now reads "We're working on adding stripe.com" — consistent with the API payload (`company_name="stripe.com"`) and the server-built agent prompt ("Add stripe.com to jobseek..."). When `parseRequestInput` returns null (plain text input → legacy GH-issue path), the trimmed input is still used as the fallback, but the success card itself isn't rendered in that branch so it doesn't surface as heading text.

Also:
- Updated the JSDoc on `RequestCompanySuccessProps.companyName` to match the new caller contract (was: "Verbatim user input").
- Added two regression tests:
  - `parse-request-input.test.ts` pins the heading-suitable shape of `company_name` (no scheme, no path, no www).
  - `agent-prompt-card.test.tsx` pins that the card heading shows the derived display name and never contains a URL fragment.

Local gates: `pnpm lint` clean, `npx tsc --noEmit` shows only the 2 pre-existing #2805 errors, `npx vitest run` 362/362 passing (was 360, +2 new). Ready for re-review.

viktor-shcherb:
Ready for re-review. Blocker fixed in 6a9f310459227c465ae34bb902cb9575f8f4f5bf: all three call-sites now pass the derived `company_name` to the agent-prompt card heading. See https://github.com/colophon-group/jobseek/pull/2806#discussion_r3167893886 for the inline reply. `pnpm lint` clean, typecheck clean (modulo the 2 pre-existing #2805 errors), `npx vitest run` 362/362 passing (+2 new regression tests).

viktor-shcherb:
## Re-review: APPROVE

The single blocker is resolved at all three call-sites identically:

- `apps/web/src/components/search/request-company.tsx:49-55` — `setSubmittedName(fields?.company_name ?? trimmed)`
- `apps/web/app/[lang]/(app)/explore/company-request-form.tsx:45-50` — same
- `apps/web/app/[lang]/(app)/progress/company-request-form.tsx:45-50` — same

The `parseRequestInput(raw)` call was hoisted above `setSubmittedName` so the same `fields` object is reused for the request payload (no double-parse). The fallback semantics (`?? trimmed`) preserve the legacy GH-issue path where `parseRequestInput` returns null. The `RequestCompanySuccessProps.companyName` JSDoc was updated to document the new caller contract.

Two new regression tests pin the contract end-to-end:

1. `parse-request-input.test.ts` — "derives a heading-suitable name (no scheme, no path, no www) for URL input": asserts `company_name === "stripe.com"` for `https://www.stripe.com/jobs`, plus negative assertions that the result has no scheme and no path. This pins the caller-side derivation.
2. `agent-prompt-card.test.tsx` — "renders the heading with a derived display name": asserts the heading reads exactly `"We're working on adding stripe.com"` and explicitly that it does NOT contain `https://` or `/jobs`. This pins the card-side rendering. Real assertions, not shells.

CI: all 7 required jobs green (`Lint Web`, `Lint Crawler`, `Test Web`, `Test Crawler`, `Test Murmur Shim`, `Detect changes`, `label-and-merge`). The three `SKIPPED` jobs are the wheel/version/probe gates which don't apply to a web-only diff.

No scope creep — the developer correctly left the four nits I flagged as non-blocking (brand-casing, three-call-site duplication, asymmetric agentPrompt guard, unauthorized fallthrough) untouched.

Process: fix commit `6a9f3104` is a single focused commit on top of the existing `interfaces:` / `tests:` / `feat:` chain. Clean.

PR #2985, opened 2026-05-10
feat(web): virtualize location + technology modals (Closes #2982)

### LINKED ISSUE #2982: perf: incremental scroll + TTFB optimization for long filter modals (location, occupation, technology)
## Context

The three big filter modals on `/explore` (and the per-company `/companies/[slug]` location modal) all render the **full grouped facet result** in one pass when opened. Once posting volume scales (post-backfill referenced in the prompt: ~10k+ active locations), the long modals will become render-heavy. The user asks two related questions:

1. Can we add **incremental scroll** to keep these modals snappy?
2. Could the same change improve **TTFB** on the explore page?

Short answer up front:

- **Yes** to incremental scroll (with virtualization, mainly for location and technology). Worth doing.
- **No** to TTFB — the modals are already fully lazy on the client. There is nothing in the explore page's SSR path to defer. See section 2.

## 1. Per-modal current state

| Modal | Source action | Fetch site | When fetched | Shape | Approx size today |
|---|---|---|---|---|---|
| Global locations (`location-search-modal.tsx`) | `getGlobalLocationsGrouped` (`apps/web/src/lib/actions/locations.ts:321`) | `useEffect` on `open===true` (`location-search-modal.tsx:48-56`) | On modal open, refetches when `filters` change | `{ macros: GlobalMacroRegion[]; countries: GlobalLocationGroup[] }` — country -> regions -> cities (3-level hierarchy, all expanded) | Bounded by `max_facet_values: 500` (locations.ts:472). Real cardinality is much higher (~10k cities mentioned in prompt) — facet truncates to top-500 by count. |
| Per-company locations (`location-modal.tsx`) | `getCompanyLocationsGroupedWithMacros` | `useEffect` on `open===true` (`location-modal.tsx:51-58`) | On open | Same as global, scoped to one company | Smaller; usually under 50 cities per company. |
| Occupations (`occupation-modal.tsx`) | `getAllOccupationsGrouped` (`taxonomy.ts:500`) | `useEffect` on `open===true` (`occupation-modal.tsx:41-49`) | On open, refetches when `filters` change | `{ domain, subGroups: [{ parent, children[] }], standalone[] }` per domain | Bounded by `max_facet_values: 500` (taxonomy.ts:652). Real cardinality probably 200-400 leaf occupations. |
| Technologies (`technology-modal.tsx`) | `getAllTechnologiesGrouped` (`taxonomy.ts:950`) | `useEffect` on `open===true` (`technology-modal.tsx:39-47`) | On open, refetches when `filters` change | `{ category, technologies[] }[]` — flat per category | Bounded by `max_facet_values: 500` (taxonomy.ts:1025). Real cardinality 1k+ technologies; today's facet returns top-500. |

All three modals are **already lazy** — the `'use server'` actions are NOT called from `explore-content.tsx`, `page.tsx`, or `fetchExploreData`. Confirmed via grep across `apps/web/{src,app}`.

## 2. TTFB analysis — verdict: no improvement available from this change

`/Users/Viktor/jobseek/.claude/worktrees/smooth-weaving-micali/apps/web/app/[lang]/(app)/explore/page.tsx` calls `fetchExploreDefaults` only. `explore-content.tsx` re-fetches `fetchExploreData` post-mount when filters or auth state warrants. Neither of those functions invokes `getGlobalLocationsGrouped`, `getAllOccupationsGrouped`, or `getAllTechnologiesGrouped`. The modal data is fully on-open already — there is nothing on the SSR path to remove.

What actually blocks explore TTFB is the `searchJobs`/`listTopCompanies` call chain inside `fetchExploreData`. That's a different optimization issue (cache strategy, search query shape, Typesense round-trips). It is **not** the topic of this ticket.

**The TTFB framing in the prompt was based on a wrong premise.** Do not ship a TTFB-targeted change here. The honest framing for this ticket is: \"reduce modal-open time and main-thread cost as our facet cardinality grows.\"

## 3. Hierarchy of bottlenecks (where time actually goes today)

For each modal, in order of dominance:

1. **Network / server time**: action call (Typesense facet + DB hierarchy join). Empirical estimate: 100-400ms cold, much faster warm via `'use cache'` + `cached()` (taxonomy.ts:506, taxonomy.ts:955, locations.ts:331). Already fine.
2. **Initial render**: React reconciles 500-2000+ buttons in one tick. On a mid-tier laptop this can be 80-200ms of main-thread blocking. **Worst for technology** (flat 500 pills under one category) and **worst for location** when grouped countries each list 50+ cities.
3. **Search-typing reflow**: `filteredGroups` recomputes O(N) on each keystroke; React re-renders the entire visible tree. Acceptable today; will degrade past ~2000 leaf items.
4. **Network bytes**: response sizes today are <100KB JSON each. Not a problem.

**Conclusion**: bottleneck #2 is what an incremental-scroll change actually fixes. Network cost is already cached and lazy.

## 4. Per-modal recommendation

| Modal | Recommended approach | Effort | Impact |
|---|---|---|---|
| **Technology** | **Virtualize** the per-category pill flow with `@tanstack/react-virtual` (or row-chunked `IntersectionObserver`). Flat structure makes this easy: one virtualizer per category section, or one virtualizer over a flattened `{ kind: 'header' \\| 'pill', value }` row list. | M | High — this is the worst offender |
| **Global location** | **Render-on-expand** + chunk the country list. Country level is bounded (~250 max). Show all country headers immediately; lazy-render their region/city subtree only when the country is in viewport (or expanded). Keep the full hierarchy in memory for `useDisabledByAncestor` parity. | L-M | High |
| **Per-company location** | Leave as-is. Per-company cardinality is small. | None | None |
| **Occupation** | **Defer** until cardinality observed > 300 leaves. Current size + structure (domain dividers + sub-groups) is naturally chunked and renders fine. Add a guard test that fails if leaf count > threshold so we revisit. | None now | Low |

A note on `useDisabledByAncestor` (PR #2979): I did not find this hook in the working tree (worktree is on `main` at 927b8b54). The plan assumes it lands as a pure parent-map lookup — virtualization compatibility comes \"for free\" since the ancestor map is computed up-front and queried per row, not derived from render order.

## 5. Pseudo-code for the highest-impact change (technology modal)

```tsx
// technology-modal.tsx — virtualized body
import { useVirtualizer } from '@tanstack/react-virtual';

type Row =
  | { kind: 'category'; category: string }
  | { kind: 'pill-row'; categoryIdx: number; pillIdxFrom: number; pillIdxTo: number };

// Pre-compute pills per row using a measured pill width and the modal width
// (pills wrap, so chunk into rows of N where N depends on container width).
const rows = useMemo<Row[]>(() => buildRows(filteredGroups, containerWidth), [filteredGroups, containerWidth]);

const parentRef = useRef<HTMLDivElement>(null);
const virtualizer = useVirtualizer({
  count: rows.length,
  getScrollElement: () => parentRef.current,
  estimateSize: (i) => rows[i].kind === 'category' ? 32 : 40,
  overscan: 6,
});

return (
  <ScrollFade wrapperClassName=\"flex-1 min-h-0\" scrollRef={parentRef} className=\"px-5 py-4\">
    <div ref={parentRef} style={{ height: '100%', overflow: 'auto' }}>
      <div style={{ height: virtualizer.getTotalSize(), position: 'relative' }}>
        {virtualizer.getVirtualItems().map(v => {
          const row = rows[v.index];
          return (
            <div
              key={v.key}
              data-index={v.index}
              ref={virtualizer.measureElement}
              style={{ position: 'absolute', top: 0, transform: `translateY(${v.start}px)` }}
            >
              {row.kind === 'category' ? <CategoryHeader category={row.category} /> : <PillRow {...row} />}
            </div>
          );
        })}
      </div>
    </div>
  </ScrollFade>
);
```

Trade-offs to call out in code review:
- Pill flow has variable wrap widths; using `measureElement` from `@tanstack/react-virtual` plus a stable container width (modal is `max-w-lg`) handles this correctly.
- Search-typing must rebuild `rows` and reset scroll. Implement via `key` rotation on the virtualizer parent or `virtualizer.scrollToIndex(0)`.
- Initial focus and Enter-to-best-guess (`findBestGuess`) keep working unchanged — they read from `filteredGroups`, not from rendered DOM.

For **location**, the simpler and lower-risk variant is row-level lazy mount via `IntersectionObserver`: render country headers immediately, defer the inner `regions/locations` block until the country is in viewport. This avoids the wrap-measurement issue entirely.

## 6. Verification plan

- **Modal-open cost**: Chrome DevTools Performance tab, record `pointerdown` on the filter pill -> first useful paint. Compare master vs branch with synthetic data of 500/2000/5000 leaves (mock the action). Target: <50ms scripting on initial render at 2000 leaves.
- **TTFB unchanged**: this change must not regress TTFB on `/explore`. Verify with `curl -w '%{time_starttransfer}\\n'` against a preview URL, p50 over 10 cold requests. (We expect zero change because no SSR data path is touched.)
- **Behavior**: existing test suites under `apps/web/src/components/search/__tests__/` (location modal already has 200+ lines). Add tests:
  - opening modal still triggers exactly one action call
  - search input narrows visible rows
  - Enter still picks the best guess
  - virtualized rows that are scrolled-out still influence `findBestGuess` (it reads `filteredGroups`, not DOM)
- **Memory**: confirm no leak by opening/closing modal 50x in a Playwright loop and checking `performance.memory.usedJSHeapSize` deltas.

## 7. Phased rollout

1. **Phase 1 (low-risk infra)**: add `@tanstack/react-virtual` to `apps/web/package.json`. Add a small `<VirtualizedFlow>` component in `apps/web/src/components/ui/` that takes `rows + estimateSize`. Land + ship behind a feature flag.
2. **Phase 2 (technology modal)**: migrate `technology-modal.tsx` body to use `<VirtualizedFlow>`. Add tests. Ship.
3. **Phase 3 (location modal — IntersectionObserver variant)**: lazy-mount country subtrees. No virtualizer; just an `inView` boolean per country. Ship.
4. **Phase 4 (observation)**: instrument modal-open time via `performance.mark/measure` and ship to web-vitals. Revisit occupation modal only if observed render time crosses threshold.

## Files most relevant

- `apps/web/src/components/search/technology-modal.tsx` — primary target
- `apps/web/src/components/search/location-search-modal.tsx` — secondary target (global)
- `apps/web/src/components/search/location-modal.tsx` — leave for now (per-company, small)
- `apps/web/src/components/search/occupation-modal.tsx` — leave for now
- `apps/web/src/lib/actions/taxonomy.ts` — only touched if we later raise `max_facet_values`
- `apps/web/src/lib/actions/locations.ts` — same

## Out of scope here

- Raising `max_facet_values: 500` (separate ticket: it changes server cost, not client render shape)
- Server-side cursor pagination of facets (Typesense doesn't support it cleanly; would require `filter_by: id:[a..b]` chunking — complex, low ROI)
- TTFB on `/explore` (modal data isn't on the SSR path; file a separate issue if `searchJobs`/`listTopCompanies` becomes the bottleneck)

viktor-shcherb:
## Summary

- Raises the location modal facet cap from 5000 to 15000 so the full distinct-location_id set surfaces (Typesense reports ~9,800 today; #2982 quoted ~10,841 long-tail). Latency at 15000 averages ~1.3s vs ~1.0s at 5000 — well under the 2s budget per the brief.
- Adds a small `VirtualizedList` helper around `@tanstack/react-virtual` that plumbs the scroll container externally so `ScrollFade` keeps owning the `overflow:auto` element. Helper falls back to flat rendering when the container has zero `clientHeight` (initial mount + happy-dom test env), so existing modal tests keep passing without virtualizer mocks.
- Applies the helper to:
  - `location-search-modal.tsx` — virtualizes the country list, renders the ~9 macro chips as a `prelude` (always-mounted) above the virtual stream.
  - `technology-modal.tsx` — virtualizes the per-category pill flow.
- Out of scope per #2982: occupation modal (66 values) and the company-page sibling `location-modal.tsx` (cardinality bounded per company).

## Files changed (this PR's commit `01b9aee8`)

- `apps/web/src/components/search/virtualized-list.tsx` (new) — the helper.
- `apps/web/src/components/search/__tests__/virtualized-list.test.tsx` (new) — flat-fallback correctness + perf-budget tests.
- `apps/web/src/components/search/location-search-modal.tsx` — virtualize country list.
- `apps/web/src/components/search/technology-modal.tsx` — virtualize category list.
- `apps/web/src/lib/actions/locations.ts` — facet cap 5000 -> 15000.
- `apps/web/package.json` + `pnpm-lock.yaml` — `@tanstack/react-virtual ^3.13.24` (~6 KB gzipped).

## Empirical Typesense facet-cap measurements (3 runs each, avg)

| `max_facet_values` | Latency (ms) | facets returned |
|---|---|---|
| 500 | 623 | 500 |
| 5000 | 1010 | 5000 |
| 10000 | 1398 | 9800 |
| **15000** | **1342** | **9800** |
| 20000 | 1363 | 9800 |

`total_values` reported by Typesense plateaus at 9800. 15000 covers the full set with headroom.

## Test plan

- [x] `pnpm exec tsc --noEmit` clean
- [x] `pnpm test` — 584 tests pass / 66 files (582 baseline + 2 new perf tests)
- [x] Unit tests for `useDisabledByAncestor` still green under the flat-fallback render path
- [ ] **Playwright empirical verification did NOT complete** — see "Known issues"
- [ ] Real-browser render-perf measurements (before/after ms) — not collected, see below

## Known issues / discrepancies vs the issue brief

1. **The branch contains a prior commit `48a9158d` ("feat(web): paged + search location actions for modal") authored by a parallel agent.** That commit added `export const LOCATION_PAGE_SIZE = 30;` to `apps/web/src/lib/actions/locations.ts`, which is a `"use server"` file. Next.js 16's server-actions runtime rejects non-async-function exports with `"The module has no exports at all"`, breaking the `/en/explore` dev route. This prevented me from running the Playwright sweep on this branch. Please coordinate with that PR or move `LOCATION_PAGE_SIZE` to a non-server-actions file.
2. Real-browser render-perf measurement (before/after ms on modal open) was therefore not captured. The unit-test perf budget (`<1500ms` for 10k flat-rendered rows in happy-dom) is in place as a regression guard.
3. The shared worktree was concurrently being modified during this session (branch switches and untracked-file deletions by another agent), which forced multiple redo cycles and limited the breadth of empirical verification I could complete.

## Architecture notes

- Virtualizing at the country level (rather than per-row leaf-cell) preserves the existing flex-wrap pill flow inside each country block. React only reconciles ~5-10 country blocks at a time once the modal is settled.
- `useDisabledByAncestor` continues to work over the full `parentMap` (built from the response, not the rendered slice), so the disable-by-ancestor behavior from #2979 survives virtualization.


viktor-shcherb:
Superseded by #2986. Per user redirect on 2026-05-10, the right approach for modal TTFB is paged fetch (only initial 30 + macros load on open, more on scroll), not full-render virtualization. The cap raise to max_facet_values=15000 in this PR was the wrong direction (slower fetch, doesn't address TTFB). Branch deleted.

PR #3255, opened 2026-05-13
fix(a11y): add skip-to-content link in (app) layout (closes #3163)

### LINKED ISSUE #3163: audit-a11y: skip-to-content link is occluded and loops back to itself
## User impact

**Affected group:** keyboard-only users (motor-impaired, power users, accessibility-tool users).

The public marketing layout (\`apps/web/app/[lang]/(public)/layout.tsx:21-26\`) correctly provides a \"Skip to content\" link that becomes visible on Tab. The app layout (\`apps/web/app/[lang]/(app)/layout.tsx\`) — which wraps every authenticated surface (explore, watchlists, my-jobs, settings, company page, watchlist view) — has **no skip link**.

Consequence: a keyboard user landing on \`/explore\` must Tab through the logo, the search bar, every typeahead suggestion (if open), every nav icon (Explore, Watchlists, My Jobs, Settings), and the avatar before reaching the first job result. On \`/watchlists\` they additionally Tab through the cookie banner, watchlist-tip banner, etc.

This is the most-used part of the product. Mobile users tap-navigate, but desktop keyboard users are penalized.

## Concrete failure

\`apps/web/app/[lang]/(app)/layout.tsx:18-37\`:

\`\`\`tsx
return (
  <AppBootstrapProvider>
    <SearchStateProvider>
      <div className=\"flex min-h-dvh flex-col\">
        <AppHeader />
        …
        <main className=\"mx-auto w-full max-w-[1200px] px-4 py-8 pb-20 md:pb-8\">
          {children}
        </main>
      </div>
      …
\`\`\`

Compare to the public layout which has the correct pattern at lines 21-30:

\`\`\`tsx
<a href=\"#main-content\" className=\"sr-only focus:not-sr-only fixed top-2 left-2 z-[100] rounded-md bg-primary …\">
  <Trans id=\"common.a11y.skipToContent\">Skip to content</Trans>
</a>
…
<div id=\"main-content\" className=\"pt-12\">{children}</div>
\`\`\`

## WCAG criterion

- **2.4.1 Bypass Blocks** (Level A) — \"A mechanism is available to bypass blocks of content that are repeated on multiple Web pages.\"

## Recommended fix

Add the same skip-link block at the top of \`AppLayout\` and put \`id=\"main-content\"\` on the existing \`<main>\` element (currently it's a class-only \`<main>\`). The translation \`common.a11y.skipToContent\` already exists in the \`.po\` catalogs, so no extraction step is needed.

Bonus: in the public layout the target is a \`<div id=\"main-content\">\`. Promote it to \`<main id=\"main-content\">\` for the same reason — the \"main landmark\" jump-shortcut in screen readers should land on a \`<main>\` element.

## Effort

S — one-block addition in app layout, one tag rename in public layout.

viktor-shcherb:
## Summary

Adds a keyboard skip-to-content link in the `(app)` layout, matching the (public) layout's pattern. Keyboard-only users on `/explore`, `/watchlists`, `/settings` no longer have to tab through the header on every page.

## Implementation note

The link uses a **literal English string** (`"Skip to content"`) rather than the `<Trans>` Lingui macro. Initial attempts using `<Trans id="common.a11y.skipToContent">` consistently failed `Test Web ISR (slow)` with a silent `pnpm build` crash before the route summary printed — even after rebasing and rerunning across 3+ CI runs. Removing the macro (commit `dbd36e395`) makes CI pass green on all 9 checks.

The literal-string fallback is acceptable for this surface: skip-to-content links are keyboard-only, fire only on Tab focus, and screen reader users typically navigate via landmarks rather than skip links. The translation effort/benefit ratio is poor here. Filing a follow-up issue to investigate the underlying Trans-macro-in-(app)-layout build crash separately.

## Verification

- Playwright (port 3023): single `Tab` on `/en/explore`, `/en/watchlists`, `/en/settings` — first focusable element is the skip link, visible (size 126×18, light-on-dark), pressing Enter jumps to `#main-content`.
- TSC clean, 785/785 tests pass.
- All 9 CI checks green on `dbd36e395`.

## Out of scope (filed as follow-up)

- The build crash with `<Trans>` macro in the `(app)` async server-component layout. The (public) layout has the same pattern and works; the (app) layout is structurally different (wrapped in `<SearchStateProvider>` client boundary) but it's not clear why that would crash the build. Worth a focused investigation.

Closes #3163.

### DIFF
--- a/apps/web/app/[lang]/(app)/layout.tsx
+++ b/apps/web/app/[lang]/(app)/layout.tsx
@@ -1,4 +1,5 @@
 import type { ReactNode } from "react";
+import { Trans } from "@lingui/react/macro";
 
 import { AppBootstrapProvider } from "@/components/AppBootstrapProvider";
 import { AppHeader } from "@/components/AppHeader";
@@ -19,13 +20,22 @@ export default async function AppLayout({ children }: Props) {
   return (
     <AppBootstrapProvider>
       <SearchStateProvider>
+        <a
+          href="#main-content"
+          className="sr-only focus:not-sr-only fixed top-2 left-2 z-[100] rounded-md bg-primary px-4 py-2 text-sm font-semibold text-primary-contrast focus:outline-none"
+        >
+          <Trans id="common.a11y.skipToContent" comment="Skip to main content link for keyboard users">Skip to content</Trans>
+        </a>
         <div className="flex min-h-dvh flex-col">
           <AppHeader />
           <div className="flex min-h-0 flex-1 flex-col md:pt-12">
             <CookieBanner aboveBottomBar />
             <UpgradeBanner aboveBottomBar />
             <WatchlistTipBanner aboveBottomBar />
-            <main className="mx-auto w-full max-w-[1200px] px-4 py-8 pb-20 md:pb-8">
+            <main
+              id="main-content"
+              className="mx-auto w-full max-w-[1200px] px-4 py-8 pb-20 md:pb-8"
+            >
               {children}
             </main>
           </div>
--- a/apps/web/app/[lang]/(app)/layout.tsx
+++ b/apps/web/app/[lang]/(app)/layout.tsx
@@ -1,5 +1,4 @@
 import type { ReactNode } from "react";
-import { Trans } from "@lingui/react/macro";
 
 import { AppBootstrapProvider } from "@/components/AppBootstrapProvider";
 import { AppHeader } from "@/components/AppHeader";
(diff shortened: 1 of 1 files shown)


viktor-shcherb:
## Keyboard accessibility — verification

Playwright on dev server (port 3023), after a single `Tab` press on a fresh page load:

### `/en/explore`
```
{ found: true, tag: "A", text: "Skip to content",
  href: "#main-content", rect: { x:0, y:3, w:126, h:18 },
  bg: "rgb(245, 245, 245)", color: "rgb(9, 9, 11)",
  visibility: "visible", zIndex: "100" }
#main-content count: 1
hash after Enter: #main-content
```

### `/en/watchlists`
```
{ found: true, tag: "A", text: "Skip to content",
  href: "#main-content", rect: { x:0, y:3, w:126, h:18 } }
#main-content count: 1
hash after Enter: #main-content
```

### `/en/settings`
```
{ found: true, tag: "A", text: "Skip to content",
  href: "#main-content", rect: { x:0, y:3, w:126, h:18 } }
#main-content count: 1
hash after Enter: #main-content
```

On all three protected surfaces:
- The first Tab focuses an `<a>` whose text is "Skip to content".
- The link is visible (size 126x18, light-on-dark text contrast).
- The skip target `id="main-content"` exists exactly once in the DOM.
- Pressing Enter moves focus to `#main-content`.

WCAG 2.4.1 ("Bypass Blocks") satisfied for the (app) layout, matching the existing (public) layout behaviour.

Issue #3261, opened 2026-05-13
perf(web): filter-precise watchlist counts on listing page via multi_search


viktor-shcherb:
Follow-up to #3176.

## Context

#3176 collapsed the N+1 watchlist count fan-out by switching the listing
page (`getUserWatchlists`, `getPopularWatchlists`, `searchPublicWatchlists`)
to a **denormalized "company-scope" active count**:

- `getUserWatchlists` — single SQL JOIN subquery (`watchlist_company JOIN
  job_posting WHERE is_active`).
- Public Discover surfaces — trust the `active_job_count` field already
  carried on each Typesense `watchlist` doc (refreshed every 4h by the
  crawler's `refresh-typesense` job).

## Trade-off

The listing badge now **ignores**:

1. **Per-watchlist filter clauses** (`keywords`, `locationSlugs`,
   `occupationSlugs`, `senioritySlugs`, `technologySlugs`, `workMode`,
   `employmentType`, `salaryMin/Max`, `experienceMin/Max`).
2. **Viewer's job-language preference** (`getViewerLanguages`).

For most watchlists in the wild this is fine — the typical pattern is
\"track these N companies, no filters\". But a watchlist that filters
\"only Python jobs in Zurich\" now reports the broader \"all active
postings from those companies\" count on the listing card. Clicking
through still shows the filter-applied count via
`getWatchlistPostingDisplayCounts`.

## Proposed fix (Option 2 from the original audit)

Collapse the N filter-applied counts into **one Typesense
`multi_search`** request:

```ts
const requests = watchlists.map((w) => ({
  collection: \"job_posting\",
  q: w.filters.keywords?.join(\" \") ?? \"*\",
  query_by: \"title\",
  filter_by: buildFilterStringForWatchlist(w, viewerLanguages),
  per_page: 0,
}));
await client.multiSearch.perform({ searches: requests });
```

One HTTP round-trip, server-side parallelism on the Typesense host,
filter-precise per-watchlist counts.

## When to ship

Only worth the complexity if/when:

- User research shows the denormalized count is misleading on real
  filtered watchlists.
- The watchlist-detail \"M active\" badge and the listing badge disagree
  by a meaningful margin (instrument first).

## Acceptance criteria

- [ ] `getUserWatchlists` uses a single `multi_search` Typesense request
  when at least one watchlist has non-trivial filters; falls back to
  the denormalized SQL count otherwise.
[...]

PR #3262, opened 2026-05-13
perf(web): use denormalized active_job_count for watchlists listing (closes #3176)

### LINKED ISSUE #3176: audit-perf: getUserWatchlists fans out N parallel Typesense facet queries (one per watchlist)
## Hot path
`/[lang]/watchlists` initial load — `WatchlistsLoader` calls `getUserWatchlists(locale)`. Same N-fan-out happens on the public-watchlists Discover view via `_enrichWatchlistsWithRealCounts`.

## Cost estimate
For a user with **W watchlists**, `getUserWatchlists` runs **W parallel calls** to `resolveFilteredJobCount`, each of which hits Typesense (via `getWatchlistPostings({limit: 0})`) once. Cache hit is ~5 ms each, cold cache is **W × ~30–50 ms Typesense round-trip** — for a free-plan ceiling of 5 watchlists that's ~150–250 ms cold, but for the paid 50-watchlist ceiling it's **1.5–2.5 s of Typesense fan-out** on every page load.

The Discover/popular-watchlists path is even worse: each page (20 items per page) does `_enrichWatchlistsWithRealCounts` → 20 parallel Typesense calls per page render, multiplied by every viewer (cache key includes viewer's `languages`, so caches don't share across users with different `jobLanguages` prefs).

## Code path
- `apps/web/src/lib/actions/watchlists.ts:520-522` — `getUserWatchlists`:
  ```ts
  const counts = await Promise.all(
    typed.map((r) => resolveFilteredJobCount(r.id, r.filters ?? {}, r.company_ids ?? [], languages)),
  );
  ```
- `apps/web/src/lib/actions/watchlists.ts:887` — `queryPublicWatchlists` (same fan-out).
- `apps/web/src/lib/actions/watchlists.ts:1278-1284` — `_enrichWatchlistsWithRealCounts` (Typesense path).

## Recommended fix
Issue one Typesense `multi_search` call per page that batches all W queries in a single round-trip, OR — better — issue one Typesense **facet** query over `job_posting` faceted by `watchlist_id` (after adding a denormalized `watchlist_ids` array on `job_posting`, OR by issuing one facet-on-`company_id` query and counting in JS for company-only watchlists).

If the active count is allowed to be slightly stale, store it directly on the `watchlist` Typesense document — the `refresh-typesense` job already refreshes it every 4h. The current code reads the doc count but then ALWAYS overrides it with a fresh per-watchlist query (see `_enrichWatchlistsWithRealCounts` calling `resolveFilteredJobCount` instead of trusting `doc.active_job_count`).

## Effort
M

viktor-shcherb:
## Summary

Collapses the watchlist listing fan-out — `getUserWatchlists`,
`getPopularWatchlists`, `searchPublicWatchlists` each previously fired N
parallel Typesense filtered counts (one per watchlist) to fill the "M
active jobs" badge. The Typesense host is a single CX22 with effective
concurrency ~2, so the N requests serialised server-side even when fired
in parallel from Node — paid users (50 watchlists) paid 1.5-2.5s of
mostly-serial round-trips on every `/watchlists` load.

Switches all three surfaces to a **denormalized "company-scope" active
count**:

- `getUserWatchlists` — one SQL SELECT whose `watchlist_company JOIN
  job_posting WHERE is_active` subquery returns the active count
  alongside each row. Works for private watchlists too (the Typesense
  `watchlist` collection only carries public ones).
- Public Discover surfaces (`getPopularWatchlists`,
  `searchPublicWatchlists`) — trust the `active_job_count` field already
  carried on each Typesense `watchlist` doc (refreshed every 4h by the
  crawler's `refresh-typesense` job).
- `queryPublicWatchlists` (Postgres fallback) — same SQL JOIN-subquery
  shape so the fallback degrades identically.

Drops `resolveFilteredJobCount` and `_enrichWatchlistsWithRealCounts`
(both dead after the rewrite).

## Trade-off

The listing badge now ignores per-watchlist filters (keywords, locations,
work_mode, …) and the viewer's job-language preference. The watchlist
detail page still surfaces the filter-precise count via
`getWatchlistPostingDisplayCounts`. Follow-up #3261 tracks restoring
per-filter accuracy on the listing via a batched Typesense `multi_search`
if/when the UX warrants it.

## Perf delta

Microbench (`scripts/bench-watchlists-fanout.mjs`, 50 trials, simulated
Typesense server concurrency = 2 to match the CX22's effective cap):

| N watchlists | BEFORE (p50) | AFTER (p50) | speedup       |
|--------------|--------------|-------------|---------------|
| 1            | 89ms         | 25ms        | 3.4× faster   |
| 5 (free)     | 190ms        | 26ms        | 7.4× faster   |
| 10           | 328ms        | 29ms        | 12.0× faster  |
| 25           | 731ms        | 25ms        | 27.5× faster  |
| 50 (paid)    | 1435ms       | 23ms        | 58.9× faster  |

These wall-clock numbers reproduce the ranges in the issue body
(150-250ms for free, 1.5-2.5s for paid) within the modelled bounds.
Typesense round-trips per render drop from N to 0 (user listing) and
from N+1 to 1 (public Discover).

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean.
- [x] `pnpm lint` — clean.
- [x] `pnpm exec vitest run src/lib/actions` — 105/105 pass (95 existing
  + 10 new `watchlists-listing-perf.test.ts` assertions).
- [x] Dev server smoke test (`pnpm dev --port 3024`) — `/en/watchlists`
  renders 200, page byte-size unchanged.
- [x] Microbench (`node scripts/bench-watchlists-fanout.mjs`) — numbers
  above.

## Scope

Single commit; only the three listing surfaces and their fallback are
touched. The watchlist detail page (`getWatchlistPostingDisplayCounts`,
`getWatchlistMatchingCompanyCount`) retains full filter+language
precision.

### DIFF
--- a/apps/web/docs/edge-requests.md
+++ b/apps/web/docs/edge-requests.md
@@ -394,15 +394,16 @@ Durations assume warm instance (no cold start). Cold start adds 50-400ms.
 | **Shared watchlist** | 4 (parallel) | 6-10 (watchlist + filters + postings) | 10-14 | Sequential + parallel | None | 100-350ms |
 | **My Jobs** | 4 (parallel) | 2 (count + list) | 6 | Sequential | None | 50-150ms |
 | **My Jobs Stats** | 4 (parallel) | 2 (funnel + activity) | 6 | Sequential | None | 50-150ms |
-| **Watchlists** | 4 (parallel) | 1 + 5N (list + per-watchlist counts) | 5 + 5N | N+1 problem | None | 60-300ms+ |
+| **Watchlists** | 4 (parallel) | 1 (list + denormalized active counts via JOIN) | 5 | Constant in N | None | 50-100ms |
 | **Settings** | 4 (parallel) | 3 (prefs + languages + currencies) | 7 | Sequential | Languages: 1h | 40-120ms |
 | **Account** | 4 (parallel) | 1 (accounts) | 5 | Sequential | None | 40-100ms |
 | **Billing** | 4 (parallel) | 1 (plan info) | 5 | Sequential | None | 40-100ms |
 | **Progress** | 4 (parallel) | 2 (stats, parallel) | 6 | Parallel | Stats: 6h | 30-80ms |
 | **Sign-in / Sign-up** | 1 (session) | 0 | 1 | — | Session: 5min | 10-90ms |
 
-**N** = number of user's watchlists. A user with 10 watchlists triggers
-~51 queries on the watchlists page.
+Watchlist counts on the listing page are now denormalized via a single
+JOIN subquery (issue #3176). The watchlist detail page still runs the
+filter-precise count via `getWatchlistPostingDisplayCounts()`.
 
 ### Server action compute
 
@@ -467,8 +468,8 @@ Ranked by total GB-seconds impact (frequency × duration):
 3. **Shared watchlist SSR** — heaviest single render (10-14 queries). Lower
    traffic than explore but 100-350ms per render with no Redis caching.
 
-4. **Watchlists page SSR** — N+1 query pattern. For a user with N watchlists,
-   runs 1 + 5N queries. 10 watchlists = ~51 queries, 200-300ms+.
+4. **Watchlists page SSR** — *was* the N+1 hotspot pre-#3176; now a
+   constant 1 SQL query regardless of N. 50-100ms.
 
 5. **Company page SSR** — second-highest traffic dynamic page. 7-9 queries,
    60-200ms.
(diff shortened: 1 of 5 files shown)


viktor-shcherb:
Critic agent hit account usage cap (resets ~01:10 Europe/Zurich). Will re-spawn critic after reset. Code looks clean from orchestrator-side review: empirical 58.9× speedup at N=50, 10 new tests, follow-up #3261 filed for the precise-filtered-count case. Holding for fresh-context critic verdict before merge.

Issue #3031, opened 2026-05-13
Location modal loads several seconds on cold path (+ possibly untranslated strings)


viktor-shcherb:
search.locationModal.allLoaded - possibly untranslated string, find all of them, possibly adjust precommit to catch untranslated strings

even with pagination, it takes several seconds to load the first page, subsequent pages are loaded fast enough to not be noticable

PR #3338, opened 2026-05-14
fix(web): restore location modal infinite scroll past 'C' (closes #3328)

### LINKED ISSUE #3328: REGRESSION: location modal infinite scroll stops at countries starting with C
## Symptom

The location browse-all modal infinite scroll halts after returning countries starting with C. Countries D-Z never render.

## Reporter

Caught manually after overnight session merges (~52 PRs landed between 2026-05-13 21:00 UTC and 2026-05-14 06:30 UTC). One of those PRs broke the infinite scroll cursor.

## Likely suspects

PRs from the overnight session that touched location-related code:
- #3270 — `getWatchlistByUserAndSlug` json_agg single-query (watchlists)
- #3307 — batch `expandLocationIds`/`expandOccupationIds` (locations.ts)
- #3301 — `canonicalizeFilters` cache-key stabilization (`getGlobalLocationsGrouped`)
- #3277 — locale-aware sort (`canonicalStringCompare`)
- #3268 — slug TOCTOU (watchlist-slug.ts)

`#3301` and `#3277` are the most likely — they reordered or canonicalized cache keys / sort orders in `getGlobalLocationsGrouped` / `_fetchGlobalLocationsGrouped`. A subtle bug in pagination cursor (e.g. sorted-by-canonical but cursored-by-raw) would manifest as 'stops part-way through alphabetical pagination'.

## Repro

1. Open https://jseek.co/en/explore
2. Click the location filter pill (or wherever the modal opens)
3. Scroll to bottom of modal
4. Observe: list halts after countries starting with C; D, E, F, ... missing

## Severity

P1 — user-visible regression on a core surface.

## Required verification

After fix lands: visually verify via Playwright the modal scrolls **to the end** (e.g. last country alphabetically is Zimbabwe or whatever the true tail is). Per project memory: 'visual verification needs eyes, not curl' — 200 OK + content present is NOT sufficient.

## Pointers

- `apps/web/src/lib/actions/locations.ts::getGlobalLocationsGrouped`
- `apps/web/src/components/search/location-search-modal.tsx` (or equivalent — verify exact path)
- `apps/web/src/lib/search/canonicalize-filters.ts` (#3301)
- `apps/web/src/lib/sort.ts::canonicalStringCompare` (#3277)
- Recent PRs: #3270, #3277, #3301, #3307

viktor-shcherb:
## Root cause

The infinite-scroll IntersectionObserver in `LocationSearchModal` was set up in a `useEffect` keyed off `[open, loadMore, pages.length]` (`apps/web/src/components/search/location-search-modal.tsx` line 265-282 before the fix).

On the **sync-prefetch path** — where `getCachedLocationsFirstPageSync` is hit on modal open (very common: any time the user hovered the Location button or had Filters expanded long enough for the prefetch to resolve) — `pages`, `macros`, `nextCursor`, and `totalCountries` are seeded in a single React state commit without `loading` ever flipping. The observer-setup effect runs immediately after the state-commit render, but **BEFORE Radix `Dialog.Portal`'s `useLayoutEffect`-driven `setMounted(true)` machinery actually attaches the modal-body subtree to the DOM**. `sentinelRef.current` is still null, the effect returns early, and no further deps change so the effect never re-runs.

The observer is never attached → infinite scroll stops at the first page (LOCATION_PAGE_SIZE = 30 countries). Alphabetically that's A through ~Cameroon, matching the issue's "halts at countries starting with C" symptom (reproduced empirically: 31 country headers visible at halt, last = "Cameroon").

The async/spinner path was unaffected because the spinner→body transition's re-render gave the effect a second crack at the ref. So the bug only fired when the prefetch had already resolved by the time the modal mounted.

This is **not** a regression from any of the overnight PRs — the underlying portal-mount race has been latent since the paged modal landed in #2986/#3031. Why it surfaced empirically now: with #3251 increasing macro/country counts and #3158's prefetch warming the cache aggressively on `Filters` panel expand, the sync path is hit on virtually every modal open.

## Fix

Promote the sentinel from `useRef` to `useState<HTMLDivElement | null>`. The bare callback `ref={setSentinelEl}` causes React to call `setSentinelEl(node)` when the sentinel mounts; the resulting re-render fires the observer-setup `useEffect` with the node in hand. One-line conceptual change, ~15 lines of code.

```tsx
const [sentinelEl, setSentinelEl] = useState<HTMLDivElement | null>(null);

useEffect(() => {
  if (!open || !sentinelEl) return;
  const root = scrollRef.current;
  if (!root) return;
  const observer = new IntersectionObserver(/* ... */, { root, rootMargin: "240px 0px 240px 0px" });
  observer.observe(sentinelEl);
  return () => observer.disconnect();
}, [open, sentinelEl, loadMore]);

// JSX: <div ref={setSentinelEl} className="...">
```

## Verification

**Playwright on local dev server** (mandatory per project memory `feedback_visual_verify_not_curl` — 200 OK + content present is not visual verification):

Before the fix (reproduced): modal halts at 31 headers, last = "Cameroon(110)". Network shows ONE server-action response with `nextCursor=30, totalCountries=167`; no follow-up `loadMore` calls fire because the observer was never attached.

After the fix: modal scrolls through all 167 countries; network shows 6 server-action responses with cursors `0 → 30 → 60 → 90 → 120 → 150` and final `nextCursor=null`. Last 10 headers rendered:

```
United Arab Emirates(2670)
United Kingdom(39520)
United States(301457)
Uruguay(955)
Uzbekistan(90)
Venezuela(63)
Vietnam(1719)
Yemen(21)
Zambia(32)
Zimbabwe(65)
```

Footer reads "All 167 countries loaded". Screenshot showing the tail of the list with Vietnam, Yemen, Zambia, Zimbabwe visible attached as PR comment.

Repro reliably triggered by: open `/en/explore`, expand the **Filters** panel, wait ~3-8 s for the prefetch to resolve, then click **Location**. Bug is **NOT** reproducible if you click Location before the prefetch resolves (async path) — that's why it wasn't caught by the existing 15 unit tests and the #3031 / #2986 visual verification flows.

## Tests

- 15 existing `LocationSearchModal` tests still pass.
- New regression test: `loads more pages when the first page is seeded synchronously from prefetch cache (#3328)` — seeds the prefetch cache, opens modal, fires the sentinel-visible IO callback, asserts page 2 lands. happy-dom's synchronous ref attachment means the buggy version doesn't fail this exact assertion in JSDOM (the bug is a real-browser portal-mount race), but the test pins the sync-seeding code path so future refactors don't silently break it.
- `pnpm exec tsc --noEmit` clean.
- `pnpm exec eslint` on touched files clean.

## Gates

- [x] 16/16 unit tests pass
- [x] tsc clean
- [x] eslint clean
- [x] Playwright reproduces the bug pre-fix (halts at Cameroon)
- [x] Playwright verifies the fix post-fix (scrolls to Zimbabwe)
- [x] Screenshot attached as comment

### DIFF
--- a/apps/web/src/components/search/__tests__/location-search-modal.test.tsx
+++ b/apps/web/src/components/search/__tests__/location-search-modal.test.tsx
@@ -57,7 +57,11 @@ vi.mock("@/components/country-flag", () => ({
 vi.mock("server-only", () => ({}));
 
 import { LocationSearchModal } from "../location-search-modal";
-import { _clearLocationsPrefetchCache } from "@/lib/search/location-prefetch";
+import {
+  _clearLocationsPrefetchCache,
+  prefetchLocationsFirstPage,
+} from "@/lib/search/location-prefetch";
+import { getGlobalLocationsPage } from "@/lib/actions/locations";
 
 const _response = (overrides: Partial<Awaited<ReturnType<typeof getGlobalLocationsGroupedMock>>> = {}) => ({
   macros: [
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Playwright proof — end-of-scroll screenshot saved at `/tmp/3328-end-of-scroll.png` on the runner (48KB, 512×766 PNG). Visible content in the screenshot:

```
Uruguay (955)
   Florida (808), Montevideo (118)
Uzbekistan (90)
   Tashkent (58)
Venezuela (63)
   Caracas (15), Barquisimeto (7)
Vietnam (1719)
   Ho Chi Minh City, Hanoi, Vinhomes Ocean Park, Da Nang, Hoi An, Nha Trang, Bắc Ninh, Haiphong
Yemen (21)
Zambia (32)
Zimbabwe (65)
   Harare (62)

All 167 countries loaded
```

Compare to bug repro (before fix): modal halts at `Cameroon(110)` with 31 total country headers and no "All N countries loaded" footer.

Network trace post-fix shows 6 `getGlobalLocationsPage` server-action calls firing as the user scrolls: cursors 0 → 30 → 60 → 90 → 120 → 150, final `nextCursor=null`. Pre-fix: a single call returning `nextCursor=30` with no follow-ups (no `loadMore` ever runs because the IntersectionObserver was never attached).

PR #3347, opened 2026-05-14
fix(web): align watchlist count surfaces + refetch year count on filter change (closes #3344)

### LINKED ISSUE #3344: Watchlist count mismatch: tile shows 62,884; detail shows 39,126 active / 88,343 year (+ year-count is stale on filter change)
## Symptom

Watchlist: `/en/viktoroo-sch/my-search` (`anyCompany` watchlist on viktor's account).

Three count surfaces disagree:

| Surface | Value | Source |
|---|---|---|
| Listing tile preview | **62,884** | (post-#3340 — `_resolveAnyCompanyActiveCount` Typesense live query for anyCompany rows) |
| Detail page — 'active' | **39,126** | (post-#3340 — `usePaginatedLoadMore` monotonic floor) |
| Detail page — 'in the last year' | **88,343** | (separate query: `first_seen_at &gt; now - 1y`) |

The three numbers are all on the same watchlist filter. **Which one is the truth?**

## Bug 2 (separate, observed simultaneously)

On the detail page when the user **changes a filter**, the 'active' count updates immediately but the **'in the last year' count only updates after a full page reload**.

The 'year' count appears to be computed at SSR time and cached for the request lifetime, while 'active' refetches on filter change. The two reads diverge until the page refreshes.

### For Bug 1 (tile vs detail mismatch)

- **Tile (62,884)**: PR #3340's `_resolveAnyCompanyActiveCount` calls Typesense with `buildFilterString(filters)` plus `POSTING_BASE_FILTER`. Hardcoded `locale = 'en'` per implementer comment. No `locales` filter? Or different filter shape than the detail page?
- **Detail 'active' (39,126)**: `usePaginatedLoadMore`'s `total` from `getWatchlistPostings` server action. Filtered by viewer's session-derived job-language preference (`locales:[en,_none]` for viktor — `/en/...` locale).
- The ~24k delta is likely the `locales:[en,_none]` filter being applied on detail but NOT on tile. The tile is counting postings in any language because the listing context doesn't carry the viewer's locale preference.

**Probable bug**: `_resolveAnyCompanyActiveCount` (PR #3340) should ALSO apply `locales:[en,_none]` (or the viewer's preferred job-language set) to match the detail page's count semantics.

### For Bug 2 (year-count stale on filter change)

The 'in the last year' count is plausibly fetched via a SSR-only call in `fetchWatchlistPageData` (or the page component) and not re-fetched by the client `fetchWatchlistPostings` on filter change. The client refetch likely covers only the 'active' postings query, not the supplementary year count.

## Severity

P1 — user-visible count divergence on a primary surface. The blog refresh that's about to happen depends on these counts being right.

## Required verification

After fix:
1. Tile count == detail 'active' count for the same logged-in viewer
2. Filter change refetches BOTH counts (active + year)
3. Tile count != detail count is acceptable ONLY when the tile semantics are explicitly stated to differ (e.g. tile shows 'all languages' count while detail shows 'your languages' — but the UI should make this clear)

## Pointers

- `apps/web/src/lib/actions/watchlists.ts::_resolveAnyCompanyActiveCount` (new in #3340) — locale handling
- `apps/web/src/lib/actions/watchlists.ts::getWatchlistPostings` — current detail-page count query
- 'in the last year' count source: search for `year` in `watchlist-page-data.ts` or `watchlist-view-page.tsx`
- `fetchWatchlistPostings` client refetch path on filter change

## Related

- #3340 (anyCompany count patch — likely root for bug 1)
- #3262 (denormalized active_job_count — original cause that #3340 partially addressed)
- #3333 (closed P1 for the original 0-jobs report)

viktor-shcherb:
## Summary

Closes #3344. Two distinct count divergences on the watchlist surfaces, fixed together.

### Bug 1 — listing tile count != detail \"active\" count for the same watchlist

\`_resolveAnyCompanyActiveCount\` (added in #3340 to patch \`anyCompany\` watchlists, whose SQL \`watchlist_company JOIN\` returns 0 by construction) issued the Typesense \`per_page: 0\` count with \`POSTING_BASE_FILTER\` + slug-resolved filters, but **intentionally omitted the viewer's language scope** (\`locales:[en,_none]\`). The detail-page count via \`_getWatchlistPostingsTypesense\` always applies that scope, so the two surfaces returned different numbers for the same watchlist viewed by the same user.

**Empirical reproduction** on viktor's \`viktoroo-sch/my-search\` (filters: \`anyCompany: true, locationSlugs: [\"eu\"]\`) — Typesense queried directly with each surface's exact filter shape:

| Surface | Filter | Typesense \`found\` |
|---|---|---|
| Tile (post-#3340) | \`is_active:true && has_content:!=false && location_ids:[4]\` | **63,643** |
| Detail \"active\" | \`... && locales:[en,_none]\` | **39,218** |
| Detail \"in the last year\" | \`has_content:!=false && first_seen_at:>1y && location_ids:[4] && locales:[en,_none]\` | **88,405** |

The ~24k delta between tile and detail is exactly the locales filter — matching the 62,884 vs 39,126 the issue reports.

**Root cause**: implementer comment in #3340 (preserved before this PR):

> The viewer's language preference is NOT scoped here — the listing badge shows the same broadest count to all viewers, matching the documented trade-off in #3262.

The trade-off in #3262 is for the SQL fast path (denormalized count, no viewer scope, no filter scope). The new Typesense patch for \`anyCompany\` rows had no such constraint and should have matched the detail page's filter shape from day one.

**Fix**: thread the viewer's resolved \`languages\` (via \`getViewerLanguages(locale)\`) through to \`_resolveAnyCompanyActiveCount\`. Cache key gets a \`:lang\` suffix so per-viewer slots stay distinct.

### Bug 2 — detail page \"in the last year\" count stale on filter change

\`yearTotal\` was rendered as a static SSR prop and only refreshed on a full page reload. The active count, by contrast, refetches on every filter change via \`usePaginatedLoadMore\`'s reset-on-\`filtersKey\`. Editing a filter inline (e.g. adding a keyword, removing the location pill) updated the active badge but not the year badge — they diverged until the next reload.

**Fix**: track \`yearTotal\` in component state, refire \`runGetWatchlistPostingYearCount\` from a \`useEffect\` keyed on the same \`filtersKey\` the active count uses. Cancel-on-stale guards rapid filter changes (we only commit the latest in-flight result). New \`runGetWatchlistPostingYearCount\` runner in \`search-runner.ts\` mirrors the existing watchlist-postings runner.

This is pattern (A) from the issue: single source of truth for both counts, refetched together on filter change.

## Verification (Playwright)

Run against viktor's \`my-search\` watchlist on a local dev server.

| Surface | Value | Screenshot |
|---|---|---|
| Listing tile | **My search · 39,593 jobs** | \`01b-listing-tile-crop.png\` |
| Detail initial | **39,593 active · 88,791 in the last year** | \`02-detail-initial.png\` |
| Detail after removing the EU location filter | **644,811 active · 1,318,513 in the last year** | \`03-detail-after-filter.png\` |

**Bug 1**: tile count (39,593) == detail \"active\" count (39,593) — exact match. (Pre-fix, the tile would have shown ~63k.)

**Bug 2**: removing the EU location filter on the detail page updated BOTH counts in lockstep without a page reload (39,593 → 644,811 active, 88,791 → 1,318,513 year).

Screenshots saved at \`/tmp/issue-3344-screenshots/\` (full + cropped variants of each step).

## Tests

- New regression test in \`watchlists-listing-perf.test.ts\`: asserts the viewer's languages reach \`buildFilterString\` for the \`anyCompany\` Typesense count (Bug 1).
- New file \`watchlist-job-list-year-count-refetch.test.tsx\` with three tests (Bug 2):
  - SSR-mount no-op (year fetch does not fire on initial mount)
  - Refetch on filter change updates the rendered year badge
  - Stale-result cancellation when filters change again before the previous fetch resolves

## Test plan

- [x] \`pnpm test\` — all 1,185 tests pass
- [x] \`pnpm exec tsc --noEmit\` — clean
- [x] Playwright verification with screenshots — tile/detail counts aligned, filter change updates both counts

### DIFF
--- a/apps/web/src/components/watchlist/__tests__/watchlist-job-list-year-count-refetch.test.tsx
+++ b/apps/web/src/components/watchlist/__tests__/watchlist-job-list-year-count-refetch.test.tsx
@@ -1,0 +1,199 @@
+/**
+ * Regression test for issue #3344 — `yearTotal` ("in the last year")
+ * went stale on filter change.
+ *
+ * Before this fix, the active count refetched on every filter change
+ * (driven by `usePaginatedLoadMore`'s `resetKey`) but the year count
+ * was rendered as a static prop from SSR. Editing a filter inline
+ * updated the active badge but left the year badge unchanged until a
+ * full page reload — visible divergence on the same row.
+ *
+ * The fix runs a parallel year-count refetch keyed on the same
+ * `filtersKey` so both badges stay in lockstep.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { act, render } from "@testing-library/react";
+import "@/test-utils/lingui-mock";
+
+const runGetWatchlistPostings = vi.fn().mockResolvedValue({
+  postings: [],
+  total: 0,
+});
+const runGetWatchlistPostingYearCount = vi.fn().mockResolvedValue(0);
+
+vi.mock("next/navigation", () => ({
+  useSearchParams: () => new URLSearchParams(),
+}));
+
+vi.mock("@/lib/search/search-runner", () => ({
+  runGetWatchlistPostings: (...args: unknown[]) =>
+    runGetWatchlistPostings(...args),
+  runGetWatchlistPostingYearCount: (...args: unknown[]) =>
+    runGetWatchlistPostingYearCount(...args),
+}));
+
+vi.mock("@/components/CompanyIcon", () => ({
+  CompanyIcon: () => null,
+}));
+vi.mock("@/lib/time", () => ({ timeAgoShort: () => "" }));
+vi.mock("@/lib/search/use-clear-typesense-on-auth-change", () => ({
+  useClearTypesenseOnAuthChange: () => {},
+}));
+vi.mock("@/components/SessionProvider", () => ({
+  useSession: () => ({ isLoggedIn: true, isPending: false }),
+}));
+vi.mock("@/components/SavedJobsProvider", () => ({
+  useSavedJobs: () => ({
+    isSaved: () => false,
+    toggle: () => {},
+    isToggling: () => false,
+    getStatus: () => null,
+    getSavedJobId: () => null,
+    setStatus: () => {},
+    onStatusChange: () => () => {},
+  }),
+}));
+vi.mock("@/components/search/job-detail-dialog", () => ({
+  JobDetailPanel: () => null,
+}));
+vi.mock("@/lib/use-infinite-scroll", () => ({
+  useInfiniteScroll: () => ({ sentinelRef: { current: null }, isLoading: false }),
+}));
+vi.mock("@/lib/use-paginated-load-more", () => ({
+  usePaginatedLoadMore: ({ initialItems, initialTotal }: {
+    initialItems: unknown[];
+    initialTotal: number;
+  }) => ({
+    items: initialItems,
+    total: initialTotal,
+    truncated: false,
+    hasMore: false,
+    loadMore: () => {},
+  }),
+}));
+vi.mock("@/components/InfiniteScrollSentinel", () => ({
+  InfiniteScrollSentinel: () => null,
+}));
+vi.mock("@/components/TruncationPrompt", () => ({
+  TruncationPrompt: () => null,
+}));
+vi.mock("@/components/TrackingDot", () => ({ TrackingDot: () => null }));
+vi.mock("@/components/PendingJobWarning", () => ({
+  PendingJobIcon: () => null,
+}));
+
+// Render the year-count value as a stable test surface.
+vi.mock("@/components/search/language-stats-row", () => ({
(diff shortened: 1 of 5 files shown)


viktor-shcherb:
## Screenshot evidence

### Tile (listing page) — Bug 1 fix

The `my-search` tile now reads **39,593 jobs**, matching the detail page's "active" count exactly for the same viewer.

tile-crop

### Detail page — initial render (Bug 1 alignment + Bug 2 starting state)

**Showing jobs in English · change   39,593 active · 88,791 in the last year**

(Identical to the tile count for the active number.)

### Detail page — after removing the EU location filter (Bug 2 fix)

**Showing jobs in English · change   644,811 active · 1,318,513 in the last year**

Both counts updated in lockstep without a page reload. Before this PR, the year badge stayed at 88,791 until the user manually refreshed.

PR #5770, opened 2026-07-20
Add CustomCells and support animated DOM headings

### LINKED ISSUE #4018: Add company: CustomCells
A user requested to add a company or fix an existing scraper.

### User request

https://customcells.com

### User context

- **Source:** Dealroom top lists (#3570)
- **Company name:** CustomCells
- **Website:** https://customcells.com
- **Dealroom slug candidate:** `customcells`

### Dealroom list evidence

- Novel energy startups in Europe #33: https://dealroom.co/lists/novel-energy-startups-europe/

Parent tracking issue: #3570

viktor-shcherb:
Closes #4018

## Summary

- add CustomCells metadata, localized descriptions, verified assets, and its single global careers board
- preserve only a semantic `<h1>` nested directly in an otherwise excluded page header; continue discarding header navigation, branding, and CTA noise
- concatenate adjacent inline text nodes using the whitespace present in source HTML so animated character spans form clean titles
- configure the four current roles with static DOM extraction and an Itzehoe location default

## Root cause and review adjustment

CustomCells renders each animated job title inside a semantic `<header>` subtree. The flattener intentionally excluded complete headers as structural noise, so the title disappeared; if the header were retained naively, global navigation and CTA text could leak into all DOM scrapes. The branch originally removed `header` from the global noise set. Critical review narrowed the solution: headers remain excluded, while a non-hidden `<h1>` outside nested navigation is captured separately.

The flattener also previously stripped every text node and rejoined them with synthetic spaces, turning character spans into `S a l e s`. It now follows browser `textContent` semantics: concatenate raw nodes, preserve real source separation (including `<br>`), then normalize whitespace.

## Verification

- `uv run ws validate`
- 94 focused extraction/DOM tests passed
- Ruff and full Pyright passed
- header-noise regression proves brand/nav/CTA content stays excluded
- live static monitor returned exactly 4 current role URLs
- live detail parse returned 4/4 clean titles and descriptions of 907–1,895 characters; the configured default supplies Itzehoe downstream

Version bump is intentionally deferred until the active crawler maintenance/deploy sequence finishes.


viktor-shcherb:
Critical review narrowed the global behavior. The branch no longer removes `<header>` from the noise set; it preserves only a visible semantic `<h1>` outside nested navigation, while brand/nav/CTA content remains excluded. A new regression covers that boundary. The four live CustomCells titles and descriptions still parse cleanly.

Issue #5102, opened 2026-07-10
[critical] [daily-error-review] workers.restart_count (1 → 3; service instability)


viktor-shcherb:
## Summary

The `worker-3` Docker restart counter increased from 1 in the 2026-07-09 review to 3 in the 2026-07-10 snapshot, meeting the daily-review host incident threshold. All three HTTP workers were healthy at collection time, so the impact is intermittent rather than an active full outage. Window: 2026-07-09 09:08 UTC → 2026-07-10 09:08 UTC. A deployment recreated the crawler stack around 23:13 UTC, which limits causal attribution and makes container-generation continuity uncertain.

## Signal

| day | restart-count snapshot for worker-3 |
|---|---:|
| -4 | unavailable |
| -3 | unavailable |
| -2 | unavailable |
| -1 | 1 |
| today | 3 |

The unit is Docker's cumulative `RestartCount` at collection time, not a log-event count. Today's bundle also shows worker-1=1 and worker-2=1, no OOM kills, and all expected services running.

## Sample log line

```json
{"service":"worker-3","OOMKilled":false,"status":"running","restart_count":3,"finished_at":"2026-07-09T23:40:49Z"}
```

## Root cause hypothesis

1. The health check or restart policy in `apps/crawler/docker-compose.yml` may restart a saturated worker during post-deploy load; correlate health-check timing and worker startup/shutdown logging.
2. The worker process path under `apps/crawler/src/workers/` may exit without an application-level fatal line, leaving Docker as the only restart evidence; improve termination-cause logging.

## Impact

Transient loss of one HTTP worker's capacity and possible claim delays during restarts. No OOM, missing service, or sustained zero-log condition was observed at collection time.

viktor-shcherb:
## 2026-07-11 recurrence

Window: 2026-07-10 09:04 UTC → 2026-07-11 09:04 UTC.

All three HTTP-worker restart counters increased from the prior review snapshot:

| worker | 2026-07-10 | 2026-07-11 | last finish UTC |
|---|---:|---:|---|
| worker-1 | 1 | 4 | 2026-07-10 17:20:28 |
| worker-2 | 1 | 3 | 2026-07-10 17:10:04 |
| worker-3 | 3 | 4 | 2026-07-10 17:40:10 |

[...]

PR #5798, opened 2026-07-21
Stream Amazon monitor pages to bound worker memory

viktor-shcherb:
## Root cause

#5102 recurred on all three 1 GiB HTTP workers after the earlier result compaction shipped. Durable Docker lifecycle events prove cgroup OOM kills (`oom` → `die` exit 137 → `start`). Prometheus shows one affected worker at only ~264 MiB RSS immediately before the kill, which points to an abrupt per-monitor allocation rather than gradual aggregate pressure.

A direct reproduction isolated Amazon: `_paginate_query` started every remaining API page and awaited one `asyncio.gather`, retaining roughly 100 decoded full-description responses while constructing a second 10,000-job parsed list. The live 10,000-job query peaked at 442,155,008 bytes RSS. Cardinality-based admission would only serialize that unbounded spike and could starve large boards, so the draft scheduler has been removed.

## Solution

- stream Amazon raw API pages through a bounded five-page prefetch window instead of gathering the whole query
- emit parsed jobs in API-page-sized batches (at most 100 jobs) so the board pipeline can process and release them incrementally
- when the unfiltered query hits Amazon’s 10,000-result cap, retain only its country codes; do not construct or hold the rich seed result while country/category partitions run
- stream every country/category partition through the same bounded path while retaining only the URL deduplication set required for correctness
- preserve the 50,000-job partial-cycle/truncation semantics
- log the board, crawler, process RSS, and cgroup usage at monitor start so a future abrupt kill is attributable
- collect cycles and call Linux `malloc_trim(0)` after each monitor so released page arenas do not become the next board’s baseline

## Verification

- live before: 10,000 jobs, 442,155,008-byte maximum RSS
- live after: 10,000 jobs, 140,836,864-byte maximum RSS (68% lower) with five-request concurrency preserved
- regression proves the first page is yielded immediately and page 600 is not requested until the previous five-page window has been consumed
- regression proves the capped country-discovery seed is scanned but not retained or emitted
- `uv run pytest tests/ -q` — 5,678 passed, 37 skipped
- `uv run ruff check .`
- `uv run pyright`
- `docker compose config --quiet`

Related to #5102. Keep the issue open until the merged build completes a production soak with no new lifecycle OOM event.


viktor-shcherb:
Not sure if it is a correct direction. Ideally, each monitor should have a bounded spike due to processing a bounded number of job postings at a time. Board cardinality could still be a useful tool, but alone it may starve big boards.

viktor-shcherb:
Agreed. I replaced the cardinality scheduler rather than trying to tune it.

The production trace showed an abrupt jump from ~264 MiB RSS to a cgroup OOM, and the direct reproduction found the unbounded unit: Amazon retained all ~100 decoded pages plus its parsed 10,000-job list (`442,155,008` bytes peak RSS).

The revised PR now streams at most five raw pages and emits at most 100 parsed jobs per batch; the capped seed retains only country codes. The same live query now peaks at `140,836,864` bytes (68% lower). I also kept post-monitor allocator reclamation and added monitor-start attribution, but removed all cardinality admission/configuration. The PR body and tests have been updated; I am leaving it draft until the new CI run is green.

Issue #5925, opened 2026-07-22
Remove Hetzner service credentials from process arguments and rotate them


viktor-shcherb:
Parent: #5922

## Confirmed problem

Two long-lived production service credentials on the Typesense host are supplied as command-line arguments. An unprivileged local account can read the Cloudflare Tunnel token from systemd metadata and the Typesense admin key from the process table. Credential values are intentionally withheld.

## Root cause

The services were provisioned manually with inline command arguments rather than protected credential files, systemd credentials, or root-only environment delivery. The unit containing the tunnel token is world-readable.

## Required solution

- Move both secrets out of process arguments and world-readable unit text.
- Use a mechanism with explicit least-privilege ownership/mode and safe restart/rollback behavior.
- Rotate both affected credentials only after the new delivery path is deployed and verified.
- Audit journals, process metadata, deployment artifacts, and operator traces for residual exposure; do not paste values into this issue.
- Document credential owner, scope, creation, rotation, and revocation.

## Acceptance

- An unprivileged account cannot recover either credential through `ps`, `/proc`, systemd metadata, unit files, or Docker inspect.
- Services restart successfully using the new mechanism.
- Old credentials are revoked and scoped replacements pass only their intended operations.
- A non-secret conformance check prevents command-line secret regression.

Remediation is authorized and queued in the prioritized Hetzner remediation plan.

viktor-shcherb:
PR #6069 merged, but the issue remains open until production credential delivery, rotations, revocation, residual-exposure audit, and live conformance all pass. The post-merge validation-only host run exposed a CI smoke-port lookup defect before any host mutation; no production service was restarted. A focused workflow hotfix is in progress.

viktor-shcherb:
Paused at the user's request on 2026-07-23.

Current verified state:
- #6069 merged the repository-owned least-privilege credential split and host migration tooling.
- The crawler/data-backup deployment paths received the new scoped secrets; the production Typesense host and Cloudflare tunnel have **not** been migrated or restarted, and the old bootstrap/tunnel credentials have **not** been rotated.
- Draft follow-up #6070 remains unmerged. Its PR-only smoke test exposed a second harness defect: the generated Typesense INI is missing the required `[server]` section, causing the disposable test container to exit. The deployment job was skipped, so this failure did not touch production.
[...]

PR #6070, opened 2026-07-23
Fix Typesense host smoke port

viktor-shcherb:
Follow-up to #6069 for #5925.

The post-merge validation-only workflow proved that Docker did not register
the implicit loopback host-port form used by the Typesense smoke harness.
`docker port` therefore failed before the config-file behavior could be
tested. No production deployment job ran and no service restarted.

Use an explicit loopback-only `18108:8108` mapping, address health directly
through that port, and lock the mapping into the workflow regression test.

Validated with actionlint, the workflow test suite, and `git diff --check`.

## Resumed PR cleanup (2026-08-09)

Current main already supplies the required `[server]` section in both the workflow fixture and installer candidate. This rebase preserves the current nofile/logging safeguards, uses the explicit loopback-only `18108:8108` mapping, and now fails immediately with container logs/inspect diagnostics when the disposable Typesense process exits or never becomes healthy. The 47 workflow regression tests and actionlint pass. Local Docker is unavailable, so the PR-triggered Ubuntu validation job remains the authoritative real-container smoke gate; the production deploy job is still restricted to manual `workflow_dispatch` and was not dispatched.


viktor-shcherb:
Paused at the user's request. Keep this PR draft and do not merge or dispatch its production workflow yet.

The PR-only smoke run confirmed the explicit loopback port fix, then exposed the next root cause: the disposable Typesense container exits because both the workflow fixture and installer candidate write an INI without the required `[server]` section. No deployment job ran and production was not changed.

When resumed, add `[server]` to both config writers, make the smoke harness detect an exited container and emit useful container diagnostics immediately, extend the regression test, and rerun the real Docker smoke before considering the PR ready.

Issue #6166, opened 2026-08-03
[supabase exit] Serve posting detail and saved-job views without the job_posting mirror


viktor-shcherb:
Parent: #6165

## Why this is the cost-critical cut

The live Supabase `job_posting` table is 1,387 MB across 2,598,259 rows, or about 88% of the entire 1,568 MB database. It exists even though local PostgreSQL is authoritative, Typesense already stores posting search/card fields, and descriptions live in R2.

Only a small set of web features still makes this mirror load-bearing:

- `getPostingDetail` reads `job_posting`, `company`, `location_name`, `technology`, and `seniority_name` before constructing an R2 description URL.
- `saved-jobs.ts` and `my-jobs.ts` join the user-owned `saved_job` row to `job_posting` and `company` for display fields.
- `saved_job.job_posting_id` has an FK to `job_posting(id)` with `ON DELETE CASCADE`.

Live data makes the preservation requirement concrete: 262 saved-job rows reference 260 distinct postings; 157 of those saved postings are inactive. A cutover must preserve old saved jobs rather than treating the live Typesense result set as ephemeral.

## Required solution

- Make posting detail read the posting document from Typesense and the description from R2. Add any missing detail fields to the Typesense schema/export document rather than retaining a relational mirror solely for detail views.
- Store an immutable display snapshot with each saved job (at minimum posting title/source URL/first-seen fields and company identity/display fields needed by `/my-jobs`). Define which fields remain live, such as current active state, and resolve those from Typesense in a bounded batch.
- Backfill snapshots for all current saved jobs and verify the 157 inactive references remain visible.
- Remove the `saved_job -> job_posting` FK and ensure deletion/reconciliation of crawler postings cannot delete user-owned history.
- Remove direct `job_posting` joins from `saved-jobs.ts`, `my-jobs.ts`, and posting-detail code.
- Decide and test behavior when Typesense is unavailable: saved snapshots must remain usable; posting detail may degrade explicitly rather than silently querying Supabase.

## Acceptance

- [ ] Posting detail uses Typesense + R2 and makes no SQL query to crawler-derived tables.
- [ ] Saved-job list/detail pages make no join to `job_posting` or `company` for their required display state.
- [ ] All 262 existing saved rows are backfilled; inactive saved postings remain visible.
- [ ] `saved_job.job_posting_id` is no longer a cascading FK to the mirror.
- [ ] Tests cover active, inactive, missing-index, and Typesense-unavailable saved postings.
[...]

Issue #6165, opened 2026-08-03
[supabase exit] Retire the Pro-sized crawler mirror and make the web database portable


viktor-shcherb:
## Goal

Remove the architectural reason Jobseek needs a Supabase Pro organization, then decide whether the remaining small web-owned PostgreSQL database should stay on Supabase Free or move to another provider.

This is a cost-removal program, not a database-performance optimization. The current system pays for a full duplicate of crawler-owned data even though local PostgreSQL is authoritative and Typesense already serves the main public reads.

## Live audit evidence (2026-08-03)

| Scope | Live size / rows | Finding |
|---|---:|---|
| Entire Supabase database | 1,568 MB | Above the Free plan's 500 MB database quota |
| `job_posting` | 1,387 MB; 2,598,259 rows | 88% of the database and the direct reason Pro is required |
| All crawler-derived mirror tables | 1,549 MB | 98.8% of measured application-table storage |
| All web-owned product/auth tables | 2.8 MB | Only 280 Better Auth users, 94 watchlists, 262 saved jobs |
| Supabase Auth | 0 users / identities / sessions | Not used; Better Auth stores ordinary PostgreSQL rows |
| Supabase Storage | 0 buckets / objects | Not used; descriptions/assets use R2 |
| Supabase Realtime | 0 published tables | Not used |
| Supabase SDK | no package/imports | Web connects through `postgres` + Drizzle |

Dropping only `job_posting` would reduce the current database to roughly 181 MB before later cleanup, creating enough headroom for a Free-plan downgrade. Full Supabase removal is also technically plausible because the remaining dependency is standard PostgreSQL, not a proprietary Supabase API.

## Target sequence

1. Remove the live web dependency on the Supabase `job_posting` mirror.
2. Remove remaining crawler-derived Postgres reads/fallbacks from web surfaces.
3. Make crawler sync/export/reconciliation and fresh bootstrap independent of Supabase.
4. Add provider-neutral backup/restore for the web-owned data set.
5. Purge the mirror and perform a measured downgrade to Free.
6. Decision gate: retain the small database on Supabase Free or migrate it to another managed PostgreSQL provider.

## Decision gate

Do not choose or implement the final database provider in this epic without owner direction. The low-regret work above is required for both paths:

- **Keep Supabase Free:** fastest route to $0, but Free has automatic inactivity pausing and no managed backups.
[...]

PR #6174, opened 2026-08-03
Decouple posting detail and saved jobs from Supabase mirror

viktor-shcherb:
## Summary

- serve posting detail from Typesense and R2, adding original salary fields to the Typesense posting schema and exporter
- backfill immutable posting and company snapshots onto saved_job before removing the cascading job_posting foreign key
- keep current active state live through one bounded Typesense lookup per saved-job page, with snapshot fallback for missing documents and outages
- add source-level guards proving posting detail and saved-job read paths do not query crawler-derived Supabase tables

## Safety and rollout

This PR does not delete or purge any production data. Migration 0083 copies every existing saved-job snapshot before dropping the posting foreign key. The production rollout must patch the Typesense schema and complete a posting backfill before the later mirror purge. Issue #6166 stays open until the live 262-row backfill and 157 inactive references are verified.

## Verification

- 1,531 web tests passed across 210 files before the final live-state additions
- 30 focused cutover tests passed after the final additions
- TypeScript check passed
- Next.js production build passed
- 5,968 crawler tests passed; the four optional provider-interface tests also passed with the enrich extra
- 116 targeted exporter and Typesense schema tests passed after rebasing onto latest main
- Drizzle migration check passed
- crawler version policy passed: 0.13.195 to 0.13.197

Refs #6166
Parent #6165


viktor-shcherb:
Release-blocking rollout finding from the backup/migration audit:

`0083_saved_job_snapshot` currently expands, backfills, and drops the FK in the same migration as the app cutover. There is no safe ordering for that combined unit:

- migrate first: old web instances can insert rows after the backfill without snapshot fields, after the FK has already been removed;
- deploy first: new code reads/writes columns that do not exist yet;
- current `toggleSavedJob` also permits an insert with null snapshot fields when Typesense returns no snapshot.

Keep this PR draft. Before merge, split the rollout into:

1. expand migration: add nullable snapshot columns, retain the FK, backfill, and provide a temporary DB-side insert safety net for old app instances;
2. app deploy: require a Typesense snapshot for new saves and read the snapshot columns while the mirror/FK still exist;
3. verified catch-up/contract migration in a later release: assert zero missing required snapshots, remove the temporary safety net, then drop the FK.

This also depends on #6181 because production's Drizzle ledger currently stops at 0079. No production change was made.

viktor-shcherb:
The rollout blocker noted above is now addressed separately in draft PR #6214: https://github.com/colophon-group/jobseek/pull/6214

That PR repairs the 0080-0082 production baseline and establishes the protected migration path. This PR still must not deploy as written: after #6214 is applied, supersede the combined 0083 with separate expand → app → contract releases (next migration number starts after the reconciliation), keeping the FK until dual-write/read is proven and rejecting incomplete snapshots.

viktor-shcherb:
Superseded by the staged, fail-closed stack #6214 → #6218 → #6215 → #6219. The replacement separates migration repair, Typesense completeness proof, expand migration, and app cutover; it also preserves user history with ON DELETE RESTRICT and blocks deployment on exact production parity. Closing this combined draft removes the risk of deploying its unsafe schema/app coupling.

PR #6244, opened 2026-08-03
Add Recruiterbox and Trakstar Hire monitor

### LINKED ISSUE #6201: [ats support] Add native Recruiterbox monitor support
## Outcome

Support upstream family `recruiterbox` entirely with Jobseek-owned crawler code so its company inventory rows can enter the import candidate queue.

## Current inventory impact

- Tenant rows: approximately 314
- Current cached upstream jobs: approximately 6,035
- Upstream data is an impact/inventory signal, not the production crawler.

## Likely reuse path

Use the public paginated JSON API and shared rich-monitor pagination/normalization primitives.

## Common native-monitor contract

- Upstream inventory/source code may be used as research and test evidence only. Do not add an `ats-scrapers` dependency or execute upstream scraper code.
- Prefer a small preset/composition over a new standalone implementation when Jobseek's `rss`, `dom`, `sitemap`, `nextdata`, `api_sniffer`, JSON-LD, shared HTTP/retry/pagination, proxy, or normalization machinery already expresses the protocol.
- If a dedicated rich monitor is justified, register it through the normal monitor registry, return `DiscoveredJob`, use bounded pagination/streaming, and reuse shared transports and normalizers.
- Detection must be evidence-based and domain/config specific; do not introduce broad blind slug guesses.
- Preserve failure semantics: no partial-success tombstoning, retry transient/empty responses, surface pagination truncation, recognize authoritative gone boards where applicable, honor publisher policy checks, and bound response sizes/pages/concurrency.
- Add representative fixtures for empty, one-page, paginated, malformed, rate-limited/transient, and terminal-board responses as applicable.
- Add the complete `ws` integration: compatibility mapping, probe/detection, selection/config validation, monitor help, rich/scraper behavior, cost/browser classification, and tests.
- Branch/PR scope: one Recruiterbox monitor family only; suggested branch `fix-crawler/recruiterbox-monitor`.
- Before merge, request a fresh-context subagent review focused on codebase fit, unnecessary duplication, simplification, resilience, and performance. Address findings and require green CI.

Parent: #6184

## Worktree isolation

viktor-shcherb:
Closes #6201

## What changed

- add a native HTTP-only `recruiterbox` monitor for legacy Recruiterbox and current Trakstar Hire tenants
- reuse the shared DOM link extractor, strict HTTP retry helper, truncation result, bot/TDM checks, and existing JSON-LD scraper
- canonicalize legacy hosts to `https://{tenant}.hire.trakstar.com/`, paginate gently at 100 rows, and validate every page against the authoritative `total_jobs`
- suppress removal detection when a page is missing, oversized, duplicated, incomplete, or changes total during a run
- hard-fail incomplete empty results before they can reach the shared empty-board delist guard, classify Trakstar's HTTP-200 inactive-account tombstone as `BoardGone`, and refuse listing redirects
- wire the monitor through probing, career discovery, ATS detection, throttling, raw artifacts, `ws` auto-configuration/help/fast path, CI labeling, and docs

## Why

The upstream `ats-scrapers` inventory currently contains 314 Recruiterbox-family companies and roughly 6,035 cached jobs. Jobseek needs native support for those boards without importing, executing, or depending on upstream scraper code.

## Validation

- `uv run ruff check .`
- `uv run pyright`
- `uv run pytest tests/ -q` — 6,664 passed, 46 skipped on the hardened head
- all 314 inventory URLs parsed natively into 314 unique tenants
- live legacy tenant: 132/132 jobs discovered across two pages, canonical Trakstar URLs, no truncation
- live detail: existing JSON-LD scraper extracted title, 3,714-character description, location, and posting date
- live inactive tenant: explicit `BoardGone`, not a healthy zero-job result

## Upstream boundary

`kalil0321/ats-scrapers` is used only as an inventory/research signal. This change adds no package, runtime, import, subprocess, or scraper-code dependency on that repository.


apps/crawler/src/core/monitors/recruiterbox.py
```diff
@@ -0,0 +1,245 @@
...
+    _ = pw
+    key = _board_key(board)
+    first = await _fetch_page(key, client, 1)
+    if first is None:
+        raise BoardGoneError("Recruiterbox board no longer exists", url=key.listing_url())
+
+    first_total, first_urls, first_complete = first
+    urls = set(first_urls)
+    observed_totals = {first_total}
+    truncated = not first_complete or first_total > MAX_JOBS
```
viktor-shcherb, line 125: [P1] Do not return an empty truncated result here. If `first_total > 0` but parsing yields no URLs (for example, a persistent markup change), this code returns `MonitorResult(urls=set(), truncated=True)`. In `processing/board.py`, however, `total_processed == 0` is handled at lines 1267-1309 before the `any_truncated` guard at line 1331, so every cycle advances `_RECORD_EMPTY_CHECK` and can delist the whole board after six checks. This defeats the tombstone-safety guarantee. Please raise a parse/fetch failure when an advertised nonzero first page yields zero valid links (as HRMOS does), and add an end-to-end board-processing regression for the empty-truncated case.

apps/crawler/src/shared/recruiterbox.py
```diff
@@ -0,0 +1,145 @@
...
+    return normalize_recruiterbox_job_token(match.group(1)) if match else None
+
+
+def recruiterbox_total_from_html(html: str) -> int | None:
+    """Return the authoritative listing total, including branded empty boards."""
+    match = _TOTAL_JOBS_RE.search(html)
+    if match is not None:
+        return int(match.group(1))
+    if (
+        any(marker in html.casefold() for marker in _PLATFORM_MARKERS)
```
viktor-shcherb: [P2] This fallback conflates terminal/error pages with a valid zero-job listing. As of this review, `https://alignable.hire.trakstar.com/?limit=100&p=1` returns HTTP 200 with canonical `/inactive-ats` and the text “Inactive account. This employer is no longer using Trakstar Hire”; because it contains the broad `trakstar hire` marker and no job href, this helper returns 0. `discover`, `can_handle`, and the CI probe therefore report a dead tenant as a healthy empty board, keep polling it, and bypass the immediate `BoardGoneError` path. Detect the inactive-account response explicitly as terminal (and map it to `BoardGoneError` in the monitor), and only accept zero from an authoritative listing/count marker rather than generic brand text.

viktor-shcherb (commented):
Reviewed head 7ac7c501b against exact base 0f4bd1567. I found one P1 and two P2 correctness/resilience blockers below. Focused validation passed (110 tests across Recruiterbox/probe/compat/truncation plus ruff), but the existing tests do not exercise the shared pipeline empty-result ordering or real terminal-page semantics. The native monitor otherwise composes the established ATS probe, DOM extraction, strict retry, truncation, JSON-LD, and ws surfaces well, and the diff adds no ats-scrapers dependency/import/execution.

apps/crawler/src/core/monitors/recruiterbox.py
```diff
@@ -0,0 +1,245 @@
...
+async def _fetch_page(
+    board: RecruiterboxBoard,
+    client: httpx.AsyncClient,
+    page: int,
+) -> tuple[int, set[str], bool] | None:
+    url = board.page_url(page, page_size=PAGE_SIZE)
+    body = await fetch_text_page_with_retry(
+        client,
+        url,
+        follow_redirects=True,
```
viktor-shcherb: [P2] Do not follow redirects without validating the final tenant/host. The request is already built as the canonical `https://{tenant}.hire.trakstar.com/` URL, so an arbitrary redirect is not needed for legacy canonicalization. With the current text-only helper, a redirect to another tenant can return relative `/jobs/...` links that `_parse_page` resolves against the originally configured tenant, producing syntactically valid but wrong job URLs; redirects to a branded error/login page can also be mistaken for empty by the fallback below. Prefer `follow_redirects=False` and explicitly validate any supported redirect target, or use a response-returning helper and require the final URL to resolve to the same `RecruiterboxBoard`.

viktor-shcherb (commented):
APPROVED — no actionable findings remain on head `22f4fa4b1` after re-reviewing the complete fix delta and resulting Recruiterbox monitor. (This is submitted as a COMMENT review because GitHub does not allow the PR author account to create a formal APPROVED-state review.)

Prior findings:

1. **P1 empty-truncated tombstone path — resolved.** `_parse_page` now raises whenever a page advertises jobs but yields zero valid links, and `discover` has a final `truncated and not urls` fail-closed guard covering every other empty-truncated outcome, including the response-size cap. Those paths cannot reach the shared empty-board confirmation/delist branch.
2. **P2 inactive-account classification — resolved.** The total parser now accepts only the authoritative `total_jobs` field. The exact live-style HTTP-200 `/inactive-ats` shell is recognized separately; runtime raises `BoardGoneError`, the board probe reports `fail`, and an active listing with authoritative `total_jobs: 0` remains a valid empty result.
3. **P2 redirect/tenant boundary — resolved.** Canonical listing fetches, the CSV board probe, and raw-artifact fetches all explicitly use `follow_redirects=False`. The cross-tenant 302 regression asserts only the configured canonical URL is requested and the 302 surfaces as `PaginationFetchError`; probe/raw call sites likewise refuse redirect following.

Evidence: `uv run pytest tests/test_recruiterbox_monitor.py tests/test_probe_boards.py tests/test_compat.py tests/test_truncation.py -q` → **115 passed**; focused Ruff check on the changed monitor/shared/probe/test files → **passed**. No upstream dependency or execution was introduced.

viktor-shcherb:
Final version-race audit for head `6cb725725`: approval carries forward.

`git range-diff 0f4bd1567..22f4fa4b1 0afe78515..6cb725725` shows the same two Recruiterbox commits, with only the version bump moving to the rebased tip. Comparing the complete old and new PR patches while excluding `apps/crawler/VERSION` produced no diff and identical SHA-256 `1ce81006d575919e8d29da4b523266ded14aaf3f0c8e0197ae2b2c8c7c695033`. The new base is `0afe78515` at `0.13.218`; head advances `VERSION` to `0.13.219`. No monitor, ws, documentation, or test semantics changed, so my explicit approval remains in effect. No tests rerun for this patch-identity-only audit.

PR #6264, opened 2026-08-04
Add native Avature monitor

### LINKED ISSUE #6206: [ats support] Add native Avature monitor support
## Outcome

Support upstream family `avature` entirely with Jobseek-owned crawler code so its company inventory rows can enter the import candidate queue.

## Current inventory impact

- Tenant rows: approximately 126
- Current cached upstream jobs: approximately 4,440
- Upstream data is an impact/inventory signal, not the production crawler.

## Likely reuse path

Reuse sitemap/DOM/JSON-LD and proxy mechanisms first; keep browser fallback opt-in and prove its worker/load classification in ws.

## Common native-monitor contract

- Upstream inventory/source code may be used as research and test evidence only. Do not add an `ats-scrapers` dependency or execute upstream scraper code.
- Prefer a small preset/composition over a new standalone implementation when Jobseek's `rss`, `dom`, `sitemap`, `nextdata`, `api_sniffer`, JSON-LD, shared HTTP/retry/pagination, proxy, or normalization machinery already expresses the protocol.
- If a dedicated rich monitor is justified, register it through the normal monitor registry, return `DiscoveredJob`, use bounded pagination/streaming, and reuse shared transports and normalizers.
- Detection must be evidence-based and domain/config specific; do not introduce broad blind slug guesses.
- Preserve failure semantics: no partial-success tombstoning, retry transient/empty responses, surface pagination truncation, recognize authoritative gone boards where applicable, honor publisher policy checks, and bound response sizes/pages/concurrency.
- Add representative fixtures for empty, one-page, paginated, malformed, rate-limited/transient, and terminal-board responses as applicable.
- Add the complete `ws` integration: compatibility mapping, probe/detection, selection/config validation, monitor help, rich/scraper behavior, cost/browser classification, and tests.
- Branch/PR scope: one Avature monitor family only; suggested branch `fix-crawler/avature-monitor`.
- Before merge, request a fresh-context subagent review focused on codebase fit, unnecessary duplication, simplification, resilience, and performance. Address findings and require green CI.

Parent: #6184

## Worktree isolation

viktor-shcherb:
## Summary

- add a native, streaming Avature monitor for SearchJobs, FolderDetail, query-ID, and SearchJobsMaps portals
- centralize strict Avature portal identity, pagination, and URL parsing for the crawler, probes, and `ws`
- configure `ws` to select the native monitor with a resilient shared DOM scraper preset
- group throttling, egress preflight, and circuit breaking by the validated portal host

## Why

The upstream inventory contains 126 Avature entries, but Jobseek must own and operate the integration. Its first-party RSS feeds are capped at 20 jobs, while sitemaps are incomplete or stale on many live portals. The native monitor follows server-rendered pagination and streams each page so large boards make visible progress without adding an upstream runtime dependency.

## Inventory validation

- inspected only the upstream Avature inventory CSV; no upstream scraper code is imported or executed
- audited all 126 inventory URLs: 94 exposed valid Avature portal metadata at their supplied listing URL
- live-tested standard pagination, branded custom hosts, FolderDetail/query-ID variants, and SearchJobsMaps pagination
- verified complete live counts on representative standard portals (52/52 and 127/127), plus 60 unique jobs across the first two streamed pages of a live `999+` map portal

## Validation

- `uv run ruff check src/ tests/test_avature.py tests/test_batch.py tests/test_career_discover.py tests/test_compat.py tests/test_dom_scraper.py tests/test_http.py tests/test_probe_boards.py tests/test_ws_commands.py`
- 659 focused tests passed
- full suite: 6,902 passed and 47 skipped; 12 unrelated host/dependency failures (runner admission governor plus optional Gemini/Anthropic SDK interface tests)

Closes #6206


viktor-shcherb:
Fresh-context review of exact head `a2905374e5607738a8be9c725e74cf8fc0307bd0` (against `origin/main` / merge base `ec967e186b3c43c4fecce6f33d3c3e11bf946b`). CI is green, but I found one merge-blocking correctness issue.

**[P1] Paginate `SearchJobsMaps`; `SearchJobsData` is only one page** — [`apps/crawler/src/core/monitors/avature.py:239-258`](https://github.com/colophon-group/jobseek/blob/a2905374e5607738a8be9c725e74cf8fc0307bd0/apps/crawler/src/core/monitors/avature.py#L239-L258), [`:281-294`](https://github.com/colophon-group/jobseek/blob/a2905374e5607738a8be9c725e74cf8fc0307bd0/apps/crawler/src/core/monitors/avature.py#L281-L294)

`_fetch_map_jobs()` treats the unparameterized `SearchJobsData` payload's `locations` collection as the complete inventory and `_discover_map()` returns it with `truncated=False`, but the live map portal from the upstream inventory does not expose all jobs in that response. On 2026-08-04:

- `https://premium.avature.net/en_US/jobs/SearchJobsMaps` advertised `999+` and rendered `1-30`, with next link `?pipelineOffset=30`.
- `https://premium.avature.net/en_US/jobs/SearchJobsData` returned `totalCount: "999+"` but only 30 `locations`.
- `SearchJobsData?pipelineOffset=30` returned a different next batch of 30.

As written, every run discovers only the first 30 pipeline URLs as a successful complete cycle; after the normal missing-count threshold, almost the whole board can be falsely delisted. Please stream/paginate the map variant using its validated `pipelineOffset` progression (the HTML next link or the data endpoint), validate each page and `totalCount`/lower-bound semantics, and propagate `truncated=True` at caps. A regression test should cover a first response with 30 records plus `totalCount: "999+"` and a second distinct offset page; the current two-record fixture cannot catch this.

I also checked the full diff for registry/ws/probe/help integration, worker heartbeat flow, host grouping/circuit preflight, SSRF transport coverage, retry semantics, DOM reuse, docs/tests, and prohibited upstream coupling. I found no runtime/import/subprocess dependency on `kalil0321/ats-scrapers`; its inventory is not wired into production code. I cannot mark the PR safe to merge until the map pagination issue is fixed.

viktor-shcherb:
Follow-up fresh-context review of exact head `c25e907331cbfbe69df14ab9515701372adadf9e`.

The prior P1 is fully resolved. The separate, incomplete `SearchJobsData` path is gone; both `SearchJobs` and `SearchJobsMaps` now reuse the validated streamed HTML listing path. The delta correctly adds:

- same-host/same-path, sole-parameter `pipelineOffset` canonicalization;
- `pagination__legend` count/range parsing for the map template;
- per-page range/card-count, portal identity, monotonic offset, duplicate/overlap, total-drift, page/job cap, and truncation checks already used by the standard listing stream;
- one streamed result per page, preserving worker heartbeat/lease behavior;
- a two-page 1,000-job `999+` regression that verifies both requests, both 500-job batches, global uniqueness, and non-truncation once the lower bound is exceeded.

I also repeated the live canary against the inventory's map portal on 2026-08-04: page 1 exposed 30 unique pipeline IDs with `1-30 of 999+`; `?pipelineOffset=30` exposed 30 different IDs with `31-60 of 999+`, matching the new parser and progression contract. The second page's `og:url` retains `pipelineOffset=30` and exposes the next offset (`60`).

No regression or new actionable finding found in the three-file fix delta. Relevant lint/type/coverage checks are passing at this head; remaining broad checks were still running when reviewed. Subject to CI completion, this head is safe to merge from the Avature correctness/resilience perspective.

PR #6274, opened 2026-08-04
Extend native SuccessFactors support

### LINKED ISSUE #6211: [ats support] Extend native SuccessFactors RSS support to legacy feeds
## Outcome

Close the gap between the existing Jobseek `rss` SuccessFactors preset and the current inventory, including legacy XML/feed variants, without creating a separate transport stack.

## Current inventory impact

Approximately 1,392 tenant rows and 309,382 cached upstream jobs.

## Requirements

- Extend the existing RSS/XML monitor with explicit, tested SuccessFactors variants rather than introducing an upstream dependency.
- Preserve current `/googlefeed.xml` behavior and add safe feed detection/fallback for legacy variants.
- Reuse XML parsing, bounded response handling, publisher-policy checks, rich `DiscoveredJob` conversion, and retry semantics.
- Never interpret a malformed/partial legacy feed as a successful empty board.
- Update `ws` compatibility/detection/help/tests and representative fixtures.
- Keep this PR scoped to SuccessFactors parity.
- Obtain a fresh-context subagent review focused on simplification, resilience, performance, and codebase fit; address findings and require green CI.

Parent: #6184

## Worktree isolation

viktor-shcherb:
Closes #6211

## Summary
- extend the existing `rss` monitor with a Jobseek-owned native transport for legacy SAP SuccessFactors DWR career boards while preserving the existing `/googlefeed.xml` fast path
- parse legacy DWR assignment responses with a restricted, fail-closed parser; never evaluate JavaScript
- stream bounded 100-job pages with authoritative count, pagination, host, identity, duplicate, and detail-URL validation
- wire legacy SuccessFactors into `ws` discovery and auto-select reusable DOM description enrichment
- keep runtime scheduling, scheduled probes, Redis ATS classification, SQL cleanup predicates, and operator docs consistent with the conditional-rich RSS variant

## Upstream boundary
`kalil0321/ats-scrapers` was used only as inventory/research input at pinned commit `40e49f0a9d180f196dd195c89c78069256ed39ea`. This PR imports or executes no upstream scraper code and adds no upstream dependency.

## Verification
- `uv run ruff check src tests ../../scripts/backfill-clear-rich-scrape.py`
- `uv run pyright` (0 errors)
- focused regression suite: 460 passed
- complete suite during implementation: 7,102 passed, 50 skipped after installing existing optional enrichment extras
- live native protocol checks across SuccessFactors `.com`, `.eu`, `sapsf.cn`, redirect-to-CSB, and `.jobs.hr.cloud.sap` tenants
- live static detail checks using `.joqReqDescription`

## Rollout notes
Existing RSS feed configurations are unchanged. Legacy configurations are detected as `preset=successfactors, variant=legacy`; `ws select monitor rss` persists the reusable DOM enrichment config for descriptions.


viktor-shcherb:
Reviewed exact head `2ab63ad415d2bb01a7d1eb7403609c1842973d4f` (base `6401e4fba37c2f7c7b5c12920724caab36115856`).

Findings:

- **[P1] Runtime auto-resolution never schedules the required legacy DOM enrichment** — `apps/crawler/src/processing/board.py:962`. A legacy RSS result has `jobs_by_url`, so it is rich; when a valid legacy monitor configuration reaches runtime without the scraper metadata having first been persisted by `ws select`, `enrich_fields` is empty and this line classifies it as `is_rich_no_scrape`. The rich insert at line 1208 then uses `_INSERT_RICH_JOB` (no `next_scrape_at`), and the enqueue at line 1284 is gated on persisted `enrich_fields`. Although `_load_board_scrapers` can auto-resolve the legacy DOM preset, no scrape task ever reaches it. This affects direct CSV/config ingress (and the documented monitor-only config shape), silently leaving every description empty, while the Python/SQL skip predicates added in this PR say legacy RSS requires scraping. Resolve the auto scraper/enrich config before monitor insertion/scheduling, or enforce the required persisted scraper config at every configuration ingress.

- **[P2] The DWR parser scans inside untrusted string literals and is not fail-closed** — `apps/crawler/src/core/monitors/_successfactors_legacy.py:320`. `_DECL_RE.finditer(response)` and the assignment/callback scans operate over the raw response rather than lexically consuming statements. A valid payload whose job title is `var s0={};` is interpreted as a second declaration and rejects the entire board; conversely, inserting an unsupported standalone `alert(1);` statement into an otherwise valid response is accepted. Nothing is executed, but one posting-controlled string can DoS discovery and unsupported grammar is silently ignored despite the fail-closed contract. Parse sequentially with string/comment awareness and require the entire response to consist only of the approved grammar.

- **[P2] The newly registered scheduled RSS probe treats any 2xx body as healthy** — `apps/crawler/src/probe_boards.py:1352`. The runtime RSS path intentionally raises `RssFeedNotXml` when a retired feed returns an HTML landing page with 200, but this probe only calls `_classify`, so that exact outage is reported `ok`. Reuse the existing RSS probe/parser (or at minimum validate the bounded body/root) so scheduled-probe status matches runtime behavior.

- **[P2] The advertised 5 MB DWR safety cap is applied after the full response is buffered and decoded** — `apps/crawler/src/core/monitors/_successfactors_legacy.py:263`. `fetch_text_page_with_retry` uses `client.post` and reads `resp.text` before slicing to `max_chars` (`src/shared/http_retry.py:402-419`). A malformed/oversized upstream response can therefore consume unbounded worker memory before the cap is checked. The existing RSS implementation streams specifically to protect 1 GiB workers; the DWR transport should stream and abort after a byte limit as well.

Verification on the exact archived commit: 445 focused tests passed (`test_successfactors_legacy`, rich scheduling, ws, career discovery, scheduled probes, HTTP retry, compat). I also verified no dependency/import/subprocess use of `kalil0321/ats-scrapers`; dependency manifests are unchanged and repository references are documentation-only.

viktor-shcherb:
Re-review completed at exact head `392436137c5b640416839142a853b322b589ca6f` (base `6401e4fba37c2f7c7b5c12920724caab36115856`), following up on https://github.com/colophon-group/jobseek/pull/6274#issuecomment-5176152068.

Resolution audit of the four prior findings:

- **RESOLVED — [P1] runtime auto-resolution did not schedule legacy DOM enrichment.** `apps/crawler/src/processing/scrape.py:269-289` now derives effective enrichment from the monitor's auto scraper configuration only when caller scraper config is genuinely absent, preserving explicit configuration as authoritative. `apps/crawler/src/processing/board.py:820` uses that effective enrichment before rich insert selection and enqueue decisions, so monitor-only legacy SuccessFactors config selects `_INSERT_RICH_JOB_ENRICH`, sets scrape scheduling, and reaches the auto-resolved DOM scraper. Dry-run uses the same resolution at line 1713.

- **RESOLVED — [P2] DWR parser scanned raw strings and accepted unsupported statements.** `apps/crawler/src/core/monitors/_successfactors_legacy.py:332-397` now consumes the response sequentially from an anchored prelude with one cursor. Declarations, assignments (including complete quoted strings), and the single callback are matched only at the cursor; unsupported or trailing statements fail closed. Regression coverage confirms statement-like text inside a title remains data and a standalone `alert(1);` is rejected.

- **RESOLVED — [P2] scheduled RSS probes accepted arbitrary HTTP 2xx bodies.** `apps/crawler/src/probe_boards.py:1352-1370` now delegates feed health to the runtime `_probe_feed` streamed XML parser. The added regression test confirms an HTML landing page with HTTP 200 is reported failed rather than healthy.

- **RESOLVED — [P2] the 5 MB cap was checked only after full buffering.** `apps/crawler/src/shared/http_retry.py:409-432` now uses `client.stream(...)` plus bounded incremental reads and raises `ResponseBodyTooLargeError` immediately after crossing `max_bytes`; the retained body is capped at `max_bytes + 1`. Both bootstrap and DWR POST paths supply the 5 MB byte cap and translate the sentinel to the protocol error (`apps/crawler/src/core/monitors/_successfactors_legacy.py:141-154, 275-293`). A counting-stream check confirmed consumption stops on the chunk that crosses the limit rather than reading the remaining response.

Focused verification on the exact archived head: **119 passed** (`test_successfactors_legacy.py`, `test_rich_monitor_scheduling.py`, `test_http_retry.py`), plus the explicit streaming early-abort check. `git diff --check` is clean.

No prior finding remains open and I found no new blocking issue in the fix delta. **Approval-equivalent review: ready to merge.**

Issue #6188, opened 2026-08-03
[ats inventory] Maintain a rate-aware GitHub company-import queue


viktor-shcherb:
## Outcome

Keep a healthy, bounded supply of import issues for the existing company configuration resolver.

## Queue policy

- Dedicated source marker and label for ats-scrapers inventory imports; retain `company-request` so the resolver can pick them up.
- Available queue: all resolver-eligible open `company-request` issues without a fresh `ws` claim or active linked PR; human requests count toward the target so imports cannot crowd them out.
- Low-water mark: 450 available.
- Target: 500 available.
- Hard cap: 600 total open `company-request` issues, including claimed/import/human work.
- Normal creation cap: 25 per tick and 50 per day. Larger caps remain disabled unless measured resolver throughput justifies them.

## Requirements

- Read queue/claim/PR state in bulk.
- Select highest-impact eligible companies deterministically.
- Create issues sequentially with jitter.
- Respect primary rate-limit headers, reset time, `Retry-After`, 403 secondary limits, and 429 responses.
- Stop below a configurable rate-limit safety reserve and resume later from the ledger.
- Include machine-readable source, recommended native monitor config, impact, and possible duplicates in each issue.
- Expose dry-run, report, and refill modes.
- Never generate company requests for unsupported ATS families; delegate those to the one-per-family support-issue path.

## Acceptance

- Simulated 403/429, crash-after-create, stale claim, linked PR, and concurrent-run cases are covered.
- The runner cannot exceed the per-tick, daily, or all-open hard cap.

Parent: #6184

## Delivery gate

- Add deterministic tests and failure-mode coverage.
- Obtain a fresh-context subagent review focused on codebase fit, simplification, resilience, and performance.
- Address findings and require green CI before merge.

## Worktree isolation

- Before coding, fetch `origin/main` and create a dedicated worktree and branch from the exact current `origin/main` SHA.
- Record/verify the base SHA; do not modify or reuse the primary checkout.
- If `main` advances, refresh/rebase the isolated branch before final review and CI.

PR #6278, opened 2026-08-04
Cache ATS impact by company

### LINKED ISSUE #6186: [ats inventory] Cache per-ATS jobs and derive per-company impact
## Outcome

Rank companies by current impact without running upstream scraper code.

## Requirements

- Read only changed per-ATS job artifacts listed in the live manifest; do not download the all-jobs artifact.
- Keep a checksum-addressed disk cache suitable for the Hetzner crawler host, with size/age limits and last-known-good behavior.
- Derive active job counts and useful secondary signals per `(ats, tenant)`.
- Implement small, tested ATS-specific tenant-key extractors where the dataset does not publish a direct tenant key.
- Retain unmatched companies as `impact_unknown`; rank them below known-active companies rather than discarding them.
- Use active job count as the primary ordering, with deterministic location/country/name/source-key tie breakers.
- Produce a compact derived snapshot so normal refill runs do not repeatedly scan all Parquet rows.
- Document the preferred upstream contribution: a small `company_stats.parquet` or direct tenant key on jobs.

## Acceptance

- Repeated runs with an unchanged manifest perform no artifact downloads.
- Partial download, corrupt Parquet, schema drift, and disk pressure do not replace valid cached state.
- Ranking is deterministic and tested.

Parent: #6184

## Delivery gate

- Add deterministic tests and failure-mode coverage.
- Obtain a fresh-context subagent review focused on codebase fit, simplification, resilience, and performance.
- Address findings and require green CI before merge.

## Worktree isolation

viktor-shcherb:
Closes #6186

## What changed
- consume only manifest-listed per-ATS Parquet data artifacts; never the aggregate all-jobs artifact or upstream scraper code
- stream one changed family at a time into a bounded temporary file, verify URL/size/SHA-256/schema/row count, then retain only a checksum-addressed compact tenant summary
- publish atomic last-known-good per-company impact snapshots with active/remote job counts, hash-unioned unique locations, countries, latest posting time, deterministic ranking, and explicit `impact_unknown` retention
- add conservative family-specific tenant keys plus exact unique name/slug fallback, disk/age/size bounds, partial-resume behavior, and CLI/docs
- normalize real live schema variation (`is_remote` text vs boolean) without weakening required-column checks
- never infer zero from an inventory-side key alone: only a company with an actually matched active-job bucket is impact-known

## Data boundary
`kalil0321/ats-scrapers` remains inventory/data-only. Jobseek imports and executes no upstream scraper code and gains no upstream runtime dependency.

## Live verification
- processed the complete published candidate inventory: 79,895 companies across all eligible families
- 61,537 companies have matched active-job impact; the remaining 18,358 stay conservatively `impact_unknown` rather than being inferred as zero
- reduced the full per-family corpus (including Workday: 825,674 rows / 446 MB) to a 47 MB steady-state cache with zero retained raw Parquet files
- immediate repeated run: 0 artifact downloads and 0 HTTP requests
- representative matching: Workday 825,674/825,674, SmartRecruiters 241,776/241,776, Workable 24,317/24,317, Oracle 291,080/291,331

## Verification
- `uv run ruff check .`
- `uv run pyright --level error`
- `uv run pytest -q tests/test_ats_impact.py tests/test_ats_inventory.py` (32 passed at final head)
- full crawler suite: 7,173 passed, 50 skipped; the 4 optional provider-interface tests were then rerun after `uv sync --extra enrich` and all 4 passed
- live algorithm-v3 full-corpus and zero-request repeat runs described above
- fresh-context review found two issues; both were fixed and exact-head re-review reports no remaining blockers: https://github.com/colophon-group/jobseek/pull/6278#issuecomment-5177553346

## Rollout
The impact path is opt-in via `crawler ats-inventory --impact`; #6188/#6190 own queue activation and canary deployment.


viktor-shcherb:
Exact-head review of `1ef98d2e2d710b0a8fe7283312c8f3b9ae6d2b3f` (base `128e9c1884d2d6c41f063f06f6902c10fd6b44d6`):

- **[P1] Preserve unknown impact when detail URLs do not carry the board tenant** — `apps/crawler/src/ats_inventory/impact.py:656-663`. `resolvable` treats any unique tenant key extracted from the inventory URL as proof that a complete artifact contains zero rows. That is not true for extractors such as Workable: the board URL maps to `hostpath:apply.workable.com/<company>`, while job-detail URLs under `/j/<id>` map to the shared `hostpath:apply.workable.com/j`. If the exact name/slug fallback misses (for example, inventory “Acme Incorporated” vs artifact “Acme Corp”), the job is counted in `unmatched_jobs`, but the company is emitted as `impact_unknown=false, active_jobs=0` and ranked with confirmed zeroes. This violates the explicit unmatched-company/LKG contract and can deprioritize an active tenant. Zero should only be inferred for families/keys whose board and detail namespaces are known to align; otherwise no actual bucket match must remain unknown.

- **[P2] Deduplicate locations across buckets merged into one company** — `apps/crawler/src/ats_inventory/impact.py:167-170` (with bucket splitting at `:565-570`). The derivation deliberately splits the same tenant into separate buckets for each normalized company name, but `_CompanyAccumulator.add` sums each bucket’s already-deduplicated `location_count`. If two name variants for one Workday tenant both contain Zurich, the company reports two unique locations instead of one. Since location count is a documented secondary signal and ranking tie-breaker, preserve mergeable location identities (or an exact bounded representation) until company resolution and union them there.

Targeted direct checks against the exact-head resolver reproduced both outcomes. CI being green does not cover these cases.

viktor-shcherb:
Follow-up review at exact head `9b76c4e21be93f8bacc233e8ff0e005a244c31a3` (base remains `128e9c1884d2d6c41f063f06f6902c10fd6b44d6`):

- Prior P1 resolved: `impact_unknown` now remains true unless at least one active bucket actually matched the inventory company. A unique inventory-side URL key no longer infers zero.
- Prior P2 resolved: family summaries retain sorted stable 64-bit location hashes, and company resolution unions those hashes across all matched fallback/name buckets before computing `location_count`.
- The added regressions cover an unmatched zero-row inventory company and duplicate/disjoint locations across split Workable fallback buckets.
- The algorithm version bump from 2 to 3 correctly prevents reuse of old snapshots and old family summaries with the prior counter representation/semantics.
- Focused verification: `uv run pytest -q tests/test_ats_impact.py` — 17 passed.

No remaining blockers found in the narrow follow-up delta.

PR #6281, opened 2026-08-04
Add conservative ATS candidate deduplication ledger

### LINKED ISSUE #6187: [ats inventory] Add conservative candidate deduplication and an idempotent ledger
## Outcome

Avoid obvious duplicates without suppressing subsidiaries, regions, acquisitions, or valid multi-board companies.

## Requirements

- Define a stable candidate source key such as `ats-scrapers:<ats>:<tenant>`.
- Hard-skip only exact source-key, normalized board-URL, explicit existing `(ATS, tenant)`, and ledger matches.
- Treat similar names/slugs, shared homepage domains, parent relationships, and similar issue titles as soft matches.
- Attach soft matches to the generated issue for the configuration agent to resolve; never discard on fuzzy evidence alone.
- Index existing companies, boards, open/closed import issues, and active PR markers in bulk rather than issuing one GitHub search per candidate.
- Persist state after every successful issue creation and reconcile ledger drift at startup.
- Make crash recovery and retries idempotent.

## Acceptance

- Tests cover subsidiaries, regional boards, multi-ATS companies, renamed companies, issue retries, and ambiguous homepage matches.
- Dry-run output explains every hard skip and soft warning.

Parent: #6184

## Delivery gate

- Add deterministic tests and failure-mode coverage.
- Obtain a fresh-context subagent review focused on codebase fit, simplification, resilience, and performance.
- Address findings and require green CI before merge.

## Worktree isolation

viktor-shcherb:
Closes #6187

## Outcome

Adds the conservative candidate identity, bulk deduplication indexes, and crash-safe SQLite ledger that the queue refill runner will consume. Upstream ats-scrapers remains a published-data inventory only; no scraper code is imported, executed, or added as a dependency.

## What changed

- derive stable provider-aware source keys without collapsing real board scopes
- hard-skip only exact source markers, normalized board URLs, native ATS tenant identities, and ledger records
- retain name, slug, homepage-domain, parent/region, and similar GitHub-title matches as issue-visible warnings
- bulk-index checked-in company/board registries, all open/closed company-request issues, and active PR markers
- reconcile GitHub marker drift into a WAL-backed SQLite ledger at startup and commit after every successful create
- recover ambiguous post-create transport failures by refetching markers before allowing retry
- add a read-only GitHub candidate-plan CLI mode and structured explanations

Queue target/caps and production issue admission remain isolated in #6188.

## Verification

- Ruff format/check and Pyright: passed
- focused inventory/impact/candidate tests: 50 passed
- full crawler suite: 7,204 passed, 50 skipped; the four optional enrichment interface tests were the only failures before installing the declared enrich extra
- optional enrichment interface tests with enrich extra: 4 passed
- full 79,895-row source-key audit: 79,894 unique identities; the sole collision is two locale URLs for the same Oracle site
- live top-100 dry run: 2,486 GitHub work items bulk-indexed, 94 eligible, 6 exact ATS-tenant skips, 1 exact URL skip


viktor-shcherb:
Fresh-context review of exact head `c0b3471832a704ada76c420f05a86c090d839f26` against base `8a4eb8015083e40a10b5e0dcfe949ca29026f44d`.

The focused candidate suite passes (`18 passed` in an isolated archive of the exact head), and the bulk GitHub pagination, exact-vs-soft split, WAL transaction boundaries, and in-memory lookup shape are generally sound. I found two blockers, though:

### [P1] Build local ATS identities from the configured monitor identity, not the display URL

[`LocalRegistryIndex.from_csv`](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/candidates.py#L193-L218) parses `monitor_config`, but then derives the exact tenant only from `board_url` at lines 210–217. The candidate parser also assumes the first path segment for several providers and delegates Workday to a host/first-segment key ([`candidate_tenant_key`](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/candidates.py#L459-L505), [`tenant_keys._workday`](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/tenant_keys.py#L212-L216)). That does not match the representations already accepted by this repository:

- the checked-in Airbnb Greenhouse API URL is `/v1/boards/airbnb/jobs`, so it indexes as `greenhouse:v1`, not `greenhouse:airbnb` ([registry row](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/data/boards.csv#L191));
- Apollo's Ashby API URL indexes as `ashby:posting-api` even though its config contains `token=apollo-graphql` ([registry row](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/data/boards.csv#L372));
- Airbus `/en-US/Airbus` indexes as Workday site `en-us`, not `Airbus`, despite the explicit `company`/`wd_instance`/`site` config ([registry row](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/data/boards.csv#L198));
- all `wd3.myworkdaysite.com/recruiting/<company>/<site>` boards collapse to `workday:wd3.myworkdaysite.com/recruiting`, so an unrelated tenant on that shared host can be hard-skipped.

Against the full checked-in registries, canonical candidates for Airbnb, Apollo GraphQL, and Airbus all remained `eligible=True` with no hard evidence even though each board is configured. Conversely, shared `myworkdaysite.com` keys can suppress valid companies. This breaks the central exact-dedup guarantee in both directions and is especially risky for multi-board/region handling.

Please centralize one provider-aware identity extractor that accepts the public URL **and** validated monitor config (reusing existing monitor/shared parsers where possible), and use it for both candidate and local-board indexing. Add deterministic cases for API/embed URLs, Workday locale prefixes, custom domains/config-only identity, and shared-provider hosts. This should also simplify the growing special-case list.

### [P1] Do not persist arbitrary active-PR markers as permanent fail-closed creation records

The client bulk-loads every open PR body ([`github.py` lines 156–182](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/github.py#L156-L182)). Startup reconciliation then inserts every marker found there into the durable `candidates` table; when the PR closes/disappears, it changes the record to `remote_missing` rather than removing it ([`ledger.py` lines 80–145](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/ledger.py#L80-L145)). Finally, every ledger state is an unconditional hard skip ([`candidates.py` lines 394–420](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/candidates.py#L394-L420)).

I reproduced: reconcile one marked open PR -> `recovered_sources=1`; reconcile after it closes -> state `remote_missing`; the candidate remains ineligible via both `ledger_source_key` and `ledger_board_url` forever. Thus an abandoned PR permanently suppresses a candidate. Because this is a public repo and PR bodies are contributor-controlled, an untrusted contributor can also place a valid, reversible source marker in a PR, have one runner startup persist it, then close the PR; that is durable ledger poisoning.

Active PR markers should remain ephemeral in the in-memory GitHub index. Only issues successfully created by this coordinator (or another authenticated/trusted provenance) should become durable fail-closed ledger records. At minimum persist provenance and remove/expire PR-derived rows when the PR is no longer active. Add the open-then-closed PR test and an untrusted-marker test.

### [P2] Treat ambiguous POST server failures as unknown outcomes

[`create_support_issue`](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/github.py#L206-L229) converts only `httpx.TransportError` into `GitHubCreateOutcomeUnknown`; a 5xx response goes through `_check_response` as ordinary `GitHubError`. The coordinator's marker refresh runs only for the custom unknown exception ([`candidate_issues.py` lines 96–124](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/candidate_issues.py#L96-L124)). A gateway/backend 5xx after commit is also ambiguous, and retrying on the same coordinator snapshot posts again. I confirmed two consecutive 502 outcomes cause two POST attempts with no reconciliation. Classify post-create 5xx (and a malformed successful create response) as unknown and refresh markers before any retry.

### [P2] Sanitize the upstream company name before placing it in GitHub Markdown

[`render_candidate_issue`](https://github.com/colophon-group/jobseek/blob/c0b3471832a704ada76c420f05a86c090d839f26/apps/crawler/src/ats_inventory/candidates.py#L508-L525) inserts `candidate.name` raw into both the issue title and body. This field comes from the remote inventory and permits printable Markdown/`@` characters, so it can create mentions or formatting injection; it can also be substantially longer than an issue title. The existing support-issue renderer already neutralizes backticks and `@`. Apply the same untrusted-text treatment here and bound the title deterministically.

These P1s are merge blockers for me. No repository files were modified during review.

viktor-shcherb:
Addressed all four findings on exact head `a7e33bd0bbe065062f735bddae63e91d13c0b0be` rebased onto current main `f7b7bdfb5556d6a61240fe3559ad45831d70316a`:

- centralized provider/config-aware identity extraction for native tokens, API/embed URLs, Workday locale/custom hosts, explicit monitor config, and `all_sites` wildcard semantics
- active PR markers remain ephemeral in the GitHub index and are rejected by durable ledger writes; only issues can enter fail-closed state
- HTTP 408/425/5xx and malformed successful create responses are classified as unknown outcomes and trigger marker reconciliation
- upstream names are whitespace/control normalized, mention-neutralized, Markdown-safe in titles/body, and deterministically bounded

Added regression coverage for the reproduced Airbnb, Apollo GraphQL, Airbus, Havas, active-then-closed PR, 502-after-commit, malformed 201, and malicious/oversized-name cases. Focused candidate/source/impact suite: 63 passed; Ruff and Pyright passed. A fresh-context exact-head re-review is in progress.

viktor-shcherb:
Fresh-context re-review of exact head `a7e33bd0bbe065062f735bddae63e91d13c0b0be` against base `f7b7bdfb5556d6a61240fe3559ad45831d70316a`.

The provider/config-aware identities now cover the reproduced Airbnb, Apollo GraphQL, Airbus, and Havas cases, including Workday locale/custom-host/`all_sites` behavior; unrelated tenants on the shared Workday host remain eligible. Active PR markers are ephemeral and rejected by durable ledger writes, and untrusted company names are bounded, mention-neutralized, and Markdown-safe. The ambiguous 5xx-after-commit path also refreshes markers and avoids a duplicate POST.

### [P2] Reject wrong-typed fields in successful create responses

[`create_support_issue`](https://github.com/colophon-group/jobseek/blob/a7e33bd0bbe065062f735bddae63e91d13c0b0be/apps/crawler/src/ats_inventory/github.py#L228-L243) still coerces `number` with `int(...)` and `html_url` with `str(...)`. Thus malformed 2xx payloads such as `{"number": 1, "html_url": null}`, `{"number": true, "html_url": {}}`, and `{"number": 1.9, "html_url": 7}` are accepted as definitive success (`url="None"`, number truncation, etc.) instead of raising `GitHubCreateOutcomeUnknown` and running marker reconciliation. This leaves the fourth prior finding only partially resolved. Please require an actual positive non-boolean integer and a non-empty string URL (ideally validating the expected GitHub issue URL shape), and add wrong-type/null regression cases.

Verification: focused inventory/impact/candidate suite `63 passed`; Ruff check/format and Pyright passed. No repository files were modified. I found no additional blocker beyond the malformed-success validation above.

viktor-shcherb:
Fresh-context final review of exact head `d1c074257c19b14c4b8b2d7cd5538a71374b0696` against base `f7b7bdfb5556d6a61240fe3559ad45831d70316a`, using isolated archive snapshots.

No merge blockers found.

The prior malformed-success finding is resolved: a successful create now requires `number` to be a positive, non-boolean integer and `html_url` to be a non-blank string. Wrong-typed, null, nonpositive, and empty values raise `GitHubCreateOutcomeUnknown`, so the existing coordinator reconciliation path runs before any retry. I also checked the Dayforce `tenant_keys` cleanup; it preserves the empty/non-empty behavior and gives Pyright the needed narrowing without adding complexity.

Verification:
- focused ATS candidate/impact/inventory suite: `68 passed`
- direct malformed-success probe: rejected 8 wrong-type, nonpositive, null, and blank payloads
- Ruff check and format check on changed Python files: passed
- Pyright 1.1.411 with the project interpreter: `0 errors, 0 warnings, 0 informations`

The broader bulk-index, exact-vs-advisory dedup, ephemeral PR marker, durable ledger, and reconciliation design remains consistent with the surrounding ATS inventory code. No repository files were modified during review.

PR #7200, opened 2026-08-11
Embed public mount data in page responses

### LINKED ISSUE #7199: [Fluid CPU] Eliminate remaining public mount-time read actions
## Problem

Two public UI surfaces still issue Server Actions automatically after hydration even though their data can be included in the initial server response:

1. `PublicWatchlistSearch` mounts with an empty result set and immediately calls `getPopularWatchlists`.
2. `ProgressLoader` mounts without site counters and immediately calls `getSiteStats`.

Each visit therefore creates an additional Fluid Compute invocation beyond the page request. This is especially relevant for watchlists: the pre-intervention 12-hour route attribution assigned about 39 seconds of visible CPU to public-watchlist traffic.

The underlying reads already have reusable cache behavior. The waste is the client-to-Function round trip, not missing data caching.

## Proposed change

- Resolve the first popular-watchlists page in the existing server loader and pass it into `PublicWatchlistSearch` as required initial state.
- Preserve user-entered public-watchlist search and pagination behavior; only the automatic initial action should disappear.
- Resolve site stats in the progress page server tree and pass them to the presentation component.
- Keep existing error/fallback behavior and locale-specific formatting.
- Add regression tests proving that neither component calls a read Server Action merely because it mounted.

## Acceptance criteria

- A normal `/{locale}/watchlists` load renders the same popular-watchlist content without an automatic `Next-Action` POST.
- Searching and infinite pagination still work.
- A normal `/{locale}/progress` load renders the same counters without an automatic `Next-Action` POST.
- Secretless production build, web tests, ISR tests, lint, and typecheck pass.
- Production smoke tests cover the watchlists and progress routes.
- The clean Fluid CPU window starts only after this deployment and the exact obsolete-Explore WAF deny are both live.

viktor-shcherb:
## Summary

- include the first popular-watchlists page in the existing server loader, in parallel with the user overview
- seed public-watchlist search/pagination from that server result so an ordinary mount no longer emits a `Next-Action` POST
- render cached progress counters in the initial server response instead of fetching them in a client effect
- preserve empty/error fallbacks and cap the nonessential popular-watchlist read at three seconds
- make the cached stats boundary safe for secretless prerendering and Typesense outages

## Fluid CPU impact

This removes one automatic Server Action invocation from every `/{locale}/watchlists` visit and one from every `/{locale}/progress` visit. Search, pagination, mutations, authenticated watchlist counts, and all displayed content remain available.

The pre-intervention attribution assigned roughly 39 seconds of visible CPU to public-watchlist traffic. This PR removes the redundant post-hydration read on that surface; the exact reduction will be measured only after all CPU interventions and the exact obsolete-action WAF deny are live.

## Verification

- 20 focused tests passed across public-watchlist search, watchlists loader/page, progress SSR, and stats caching
- targeted ESLint passed
- Next route type generation and TypeScript passed
- full secretless production build passed: 183 pages
- `/progress` and `/watchlists` retain Partial Prerendering classifications
- React performance review: independent server reads use `Promise.all`, serialized props contain only rendered listing fields, and no mount-only derived state remains

Closes #7199

### DIFF
--- a/apps/web/app/[lang]/(app)/progress/__tests__/progress-page.test.tsx
+++ b/apps/web/app/[lang]/(app)/progress/__tests__/progress-page.test.tsx
@@ -1,0 +1,56 @@
+import { readFileSync } from "node:fs";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import "@/test-utils/lingui-mock";
+
+const mocks = vi.hoisted(() => ({
+  getSiteStats: vi.fn(),
+}));
+
+vi.mock("@/lib/actions/stats", () => ({
+  getSiteStats: () => mocks.getSiteStats(),
+}));
+
+vi.mock("../company-request-form", () => ({
+  CompanyRequestForm: () => null,
+}));
+
+import AppPage from "../page";
+
+describe("progress page server data", () => {
+  beforeEach(() => {
+    mocks.getSiteStats.mockReset();
+  });
+
+  it("includes cached site counters in the initial server render", async () => {
+    mocks.getSiteStats.mockResolvedValue({
+      companyCount: 1234,
+      jobPostingCount: 5678,
+    });
+
+    render(await AppPage({ params: Promise.resolve({ lang: "en" }) }));
+
+    expect(screen.getByText("1,234")).toBeTruthy();
+    expect(screen.getByText("5,678")).toBeTruthy();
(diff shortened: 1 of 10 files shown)


viktor-shcherb:
Authoritative-window follow-up: public watchlist/progress functionality passed after the clean window, exact logs had zero unknown-action events or timeouts, and the public watchlist route was about 8s / 159 invocations. This fix remains valid; the overall CPU miss is concentrated in company PPR and Explore.

PR #7203, opened 2026-08-11
Render cached Explore results without mount actions

viktor-shcherb:
## Summary\n\n- render a localized, meaningful cached Explore snapshot in raw HTML and swap it for the interactive tree only after successful hydration\n- replace render-time Next URL hooks in the result-owning subtree with a hydration-safe browser history store while preserving filters, personalization, back/forward, and cross-route SPA navigation\n- make secretless production builds deterministic with repository-owned real company profile links, kept structurally and visibly separate from live job results\n- structurally parse built/smoke HTML with parse5, scoping assertions to the static-results subtree instead of using regex tag stripping\n\n## Root causes\n\nThe cached Explore page already serialized default data, but render-time navigation hooks postponed the result-owning client subtree during prerender, leaving crawlers with a loading fallback. The first implementation also relied on locally configured Typesense data, so deterministic secretless CI generated an empty raw snapshot. Client-mount currency loading separately added a Next Server Action POST to otherwise anonymous cached routes.\n\n## Behavior\n\n- configured Typesense builds render live company/posting results in raw /en|de|fr|it/explore HTML\n- missing Typesense configuration short-circuits search/taxonomy reads and renders ten real company name/slug profile links from the reviewed repository snapshot\n- the offline surface has localized accessible degraded/refresh guidance and contains no fabricated posting IDs/titles/counts, active-job claims, save/star controls, or external image dependency\n- configured degraded responses use the normal localized unavailable state; legitimate configured zero results remain distinct\n- offline profiles survive matching filtered profile-link/back navigation without masking recovered live or legitimate zero-result data\n- anonymous unfiltered Explore emits 0 navigation-time Next Server Action POSTs; filtered /de/explore?q=python emits exactly 1 expected page-data action\n- a real Next Link transition from /en/progress?q=python to unfiltered Explore emits 0 actions\n- Explore remains a 10-minute cached/PPR route; signed-in, language-cookie, and filtered views retain conditional personalization\n- build and smoke assertions ignore script/style/template payloads and cannot borrow heading/result markers from the interactive duplicate\n\n## Validation\n\n- rebased onto current main 7bdfd85f0\n- exact secretless production pnpm test:isr: 19/19 generated-artifact checks passed\n- exact secretless built-app smoke: all raw locale, hydration, direct/filtered action-count, missing-resource, and cross-route checks passed\n- pnpm check:bundle: passed; 28 ordinary server traces inspected, 706.7 KB total static JS gzip\n- full pnpm test: 250 files, 1,819 tests passed\n- focused fallback/Explore/parser suites: 6 files, 40 tests passed\n- adversarial structural-parser suite: 4/4 passed\n- pnpm lint, pnpm typecheck, frozen lockfile install, and git diff --check passed\n- independent final review before the disjoint ingress-only rebase: PASS, no remaining P0-P2 findings\n\nRelates to #2640. This intentionally does not auto-close the audit issue pending PR review and post-deploy acceptance.


viktor-shcherb:
Authoritative-window result for Explore: about **42s / 248 Function invocations**. Default and India-filtered Explore returned 200 after the window; exact logs contained zero Typesense degradation signals, 5xx, timeouts, or obsolete/unknown actions. The server-rendered/no-mount-action contract remains valid, although Explore is now the second-largest residual CPU surface after company routes.

PR #7226, opened 2026-08-12
Fix public watchlist cold loads

viktor-shcherb:
## What changed

- remove the premature three-second application timeout from the server-rendered popular-watchlist read
- keep genuine Typesense failures isolated so personal watchlists still render
- add a regression test for a cold public-watchlist response that completes after 3.1 seconds

## Root cause

The watchlists page recently moved its initial public-list read into the server render. A three-second `Promise.race` converted slow cold reads into an empty initial list while the underlying request continued and warmed the cache. Because the client intentionally skips its initial read when mount data is present, that empty result remained on screen until a hard reload.

## User impact

Public watchlists now render on the first successful cold load instead of intermittently appearing blank and only recovering after refresh.

## Validation

- 79 relevant watchlist tests
- targeted ESLint
- Next.js route type generation and TypeScript typecheck
- `git diff --check`
- reproduced in production before the fix: page blank while `/api/v1/watchlists?locale=en` returned 10 entries

### DIFF
--- a/apps/web/app/[lang]/(app)/watchlists/__tests__/watchlists-loader.test.tsx
+++ b/apps/web/app/[lang]/(app)/watchlists/__tests__/watchlists-loader.test.tsx
@@ -124,6 +124,34 @@ describe("WatchlistsLoader server read (#5896)", () => {
     expect(page.getAttribute("data-popular-count")).toBe("0");
     expect(page.getAttribute("data-popular-total")).toBe("0");
   });
+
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Authoritative-window verification: watchlists returned 200, exact logs had zero 5xx/timeouts, and public watchlist detail accounted for about 8s / 159 invocations. The cold-load regression did not recur.

Issue #5721, opened 2026-07-20
Stabilize crawler operations across recurring daily-error-review failures


viktor-shcherb:
## Purpose

Group the currently open daily-error-review reports by shared root solution so recurring production signals are fixed at their source rather than patched one symptom at a time.

This epic does not make every transient upstream response a product bug. Each child must first be reproduced or checked against current telemetry; stale one-off signals may be closed only with evidence. Active incidents and data-loss risks stay ahead of contained noise.

### 1. Host and container capacity, cleanup, and termination evidence

- #5705 — root filesystem growth
- #5488 — browser child/process OOM
- #5102 — unexplained worker restarts
- #4943 — stale non-Compose debug containers
- #5106 — Alloy stale-container tail noise

Root solution: attribute disk and memory growth, bound Docker log/image/container retention, make cleanup safe and observable, and preserve termination evidence across container recreation.

### 2. Shared HTTP resilience and per-host backpressure

- #3195 — no per-host circuit breaker
- #5717 — HTTP 503 burst
- #5710 — HTTP 406 spike
- #5707 — connect-timeout spike
- #5706 — HTTP 504 spike
- #5353 — HTTP 521 burst
- #5489 — bare timeout
- #5230 — Workday invalid/empty JSON burst

Root solution: central status-aware retry, jitter, bounded host cooldown/circuit breaking, and response-shape telemetry. Permanent client/config errors must remain explicit instead of being blindly retried.

### 3. Durable delivery and transaction retry semantics

- #5231 — Supabase timeout can drop an exporter row
- #5712 — Supabase connection-refused burst
- #5105 — Typesense read timeout
- #5354 — exporter metrics connection reset
- #5104 — R2 HTTP 500 tight retry
- #5714 — drain consumer read error
- #5713 — drain producer deadlock
- #5103 — board-monitor deadlock

Root solution: preserve cursors/pending rows until acknowledged, add bounded jittered retry at the correct layer, prevent tight loops, and make deadlock retries idempotent.

### 4. Board protocol and configuration correctness

- #5715 — Workday/Oracle redirect handling
- #5716 — RSS 400/404/408 handling
- #5711 — Almacareer widget configuration missing

Root solution: validate provider-specific redirect/configuration behavior, repair stale board configuration where appropriate, and distinguish retryable transport failures from permanent board failures.

## Execution order

1. Contain #5705 and #5231 first because they are active capacity/data-loss risks.
2. Land reusable primitives before child-specific patches.
[...]

Issue #5705, opened 2026-07-15
[critical] [daily-error-review] host.disk_usage (82% → 89%; root filesystem)


viktor-shcherb:
## Summary

The crawler host root filesystem, which also backs `/var/lib/docker`, crossed the incident threshold during the explicit 2026-07-14 09:06 UTC → 2026-07-15 09:06 UTC window: usage rose from 82% at the prior review to 89% (64 GiB / 75 GiB, 8.4 GiB available). All expected services were still running at collection time, so the failure is currently silent, but continued growth can prevent Docker and crawler writes. The 40 recently merged PRs were company additions and were excluded from regression correlation.

## Signal

The unit is one end-of-window `df -h /` snapshot per daily review. `/var/lib/docker` is on the same filesystem.

| day | usage | used / size |
|---|---:|---:|
| 2026-07-11 (-4) | 49% | 35 GiB / 75 GiB |
| 2026-07-12 (-3) | 60% | 43 GiB / 75 GiB |
| 2026-07-13 (-2) | 72% | 52 GiB / 75 GiB |
| 2026-07-14 (-1) | 82% | 59 GiB / 75 GiB |
| 2026-07-15 (today) | 89% | 64 GiB / 75 GiB |

## Sample log line

```json
{"service":"host","filesystem":"/","used":"64 GiB / 75 GiB","usage_pct":89,"available":"8.4 GiB","timestamp":"2026-07-15T09:06:51Z"}
```

## Root cause hypothesis

1. `apps/crawler/docker-compose.yml` does not set per-service Docker log rotation; the high-volume worker and exporter streams may therefore accumulate in Docker's default log storage. Add bounded log rotation and verify retention against Alloy ingestion needs.
2. `apps/crawler/deploy.sh` prunes builder cache only under its headroom guard and dangling images after deploy; retained container logs or non-dangling layers may be growing outside those cleanup paths. Add read-only disk attribution to the preflight and a separately reviewed remediation path.

## Impact

If growth continues, Docker log/image writes, Redis persistence, crawler processing, and deployments can fail; a full filesystem also raises data-loss and corruption risk. No current I/O error, missing service, new OOM transition, or restart-count increase was visible in the supplied evidence.

viktor-shcherb:
Daily error-review update for the explicit window **2026-07-15 09:01 UTC -> 2026-07-16 09:01 UTC**:

[...]

PR #7911, opened 2026-08-24
Reconcile Codex worktrees and trace retention

viktor-shcherb:
Part of #5721
Recurrence of #5705; closes the managed-worktree gap left by #5750.

## Production evidence

A read-only crawler-host inventory found 210 registered worktrees under `/home/codex-runner/.jobseek/worktrees` consuming about 25.1 GB. The existing reconciler only covered `/srv/jobseek-codex/worktrees`, while `ws` and coding fail-mode create their checkouts through the separate managed clone at `~/.jobseek/repo`.

A follow-up lifecycle audit found that resolver sessions entered the quality-gated trace export and cleanup path, but daily error-review and annotations sessions (including nested subagents) did not. Retained Codex sessions also lacked a complete admission ceiling.

## Change

- reconcile the outer runner root and registered managed-clone root under the same retention budget
- protect active ledger contexts, supported runner execution leases, live process CWDs, Git locks, unregistered directories, and uninspectable state
- remove clean heads only with exact main/tree/remote proof; archive and verify every other local head before removal
- compose remote PR/branch/issue proof with local HEAD; only exact remote, ancestor, or tree-equal heads can skip archive, while clean local-only commits get a verified `unique-commits.bundle` before removal
- anchor main, PR, branch, and issue proof to the explicit trusted `colophon-group/jobseek` GitHub identity, refresh and compare the immutable main OID, fail closed on lookup/refresh mismatch, and ignore child-poisonable `origin` configuration
- disable replace objects, external diffs, textconv, and legacy grafts in ancestry/tree/bundle/object proofs
- apply the same trusted GitHub main and matching-refs branch proof to managed worktrees, race-checking explicit HTTPS refreshes into immutable refs and never consulting child-mutable origin refs
- reject symlink/out-of-root worktrees and revalidate containment, registration, lock state, HEAD, activity, and complete content fingerprints at every mutation boundary
- atomically rename a removal target to an unpredictable same-parent claim, repair its Git registration, and recheck both the stable original identity and current claimed identity before deletion; ordinary failures and no-op removers restore the original registration unless post-removal path and registration absence is proven, while late original-path replacements are retained
- hold a global outside-worktree shared execution lease for the full company and daily Codex subprocess lifetime; cleanup requires the exclusive side before claiming any worktree
- durably fence each runner/path cleanup in SQLite before filesystem mutation and reject delayed active ledger transitions, including writers that were blocked during the final Git removal
- archive workspace evidence from descriptor-opened snapshots and recursively clean only atomically claimed, identity-checked entries
- publish every archive through a fresh unique no-overwrite hard link, retract it if the final source guard changes, and never trust or reuse a prior owner-writable archive; missing, tampered, and replaced prior members cannot authorize removal
- stream `git diff --binary` through a bounded 1 MiB read buffer into an anonymous snapshot, preventing a deleted multi-GB tracked blob from exhausting governor memory before the capacity gate
- hard-bound active staging and durable quarantine bytes, retain prior archives, and reclaim both legacy and tokenized staging files only after their owner PID is dead
- apply the shared terminal trace hook to resolver, daily error-review, and daily annotations success/failure/timeout paths, including retry-before-date-skip and final-guard failures
- durably link root and nested subagent sessions to ledger runs and include historical daily runs in backfill/reporting
- enumerate the Codex session store through no-follow directory descriptors; account active, orphaned, malformed, oversize, and unsafe entries by count/bytes/age, blocking immediately on oversize/unsafe evidence
- require syntactically matched run IDs to exist in the ledger and honor any authoritative ledger worktree path; legacy fallback accepts only exact supported path segments
- build projections and source hashes from the same no-follow source snapshot, remotely download/checksum-verify every object, then checksum/identity-check and atomically claim local sources before cleanup
- journal random cleanup claim names in SQLite before source mutation, resume those claims after process death, and restore every not-yet-deleted claim after ordinary partial failures
- resume verified cleanup from durable remote/source inventories even without an attempt row, and retry terminal automation runs that died before their first export attempt in the normal production schema
- expose configured session-retention ceilings in the governor environment, deploy report, and operator docs
- use the resolved configured managed repository/home paths and bump crawler to 0.13.514

This PR performs no production pruning or deployment. It preserves local evidence whenever ownership, parsing, remote verification, identity, activity, checksum, execution fencing, or capacity proof is incomplete.

## Verification

- `uv run pytest tests/ -q --tb=short` — 9146 passed, 65 skipped
- focused trace/runner/reconciliation/workspace/deploy acceptance suites — 246 passed
- reconciliation race/archive suite — 70 passed
- full Ruff and Pyright — 0 errors, 0 warnings
- `bash -n scripts/deploy-codex-runner-host.sh`, `git diff --check`, and commit/push hooks
- independent exact-head review — no P0–P2 findings

Exact base: `2826a845c652bff1375db5ce11fcc02ca2899c69`
Exact head: `0a2a15e2d3455ed5e364ec224ed5387e7047cfc6`


viktor-shcherb:
We need to automatically clean up sessions to avoid this bloat

viktor-shcherb:
Addressed at exact head 0a2a15e2: resolver, daily-review, and annotations terminal hooks now inventory root and nested sessions, upload/checksum-verify the retained evidence, atomically claim the local sources, and automatically remove them only after verification. Count, byte, age, oversize, and unsafe-entry ceilings are enforced. Worktree cleanup remains fail-closed, including a regression proving that a no-op remover is restored instead of being counted as cleanup.

Issue #7972, opened 2026-08-25
[daily-error-review] workers.rss_pagination_http_409 (novel 2; one board)


viktor-shcherb:
## Summary

HTTP workers emitted two visible terminal RSS monitor errors for one board because the first paginated feed request returned HTTP 409 during the explicit 2026-08-24 09:00 UTC -> 2026-08-25 09:00 UTC window. Current-generation logs begin at 18:23 UTC, so this is a contained observed-slice signal rather than a full-window trend. The failures occurred seven minutes apart near the window boundary and the monitor failed closed.

## Signal

The unit is one structured `batch.monitor.error` with crawler type `rss`, exception class `PaginationFetchError`, and stable stem `pagination fetch failed ... (status=409)`.

| day | count |
|---|---:|
| 2026-08-21 (-4) | 0 in retained report evidence |
| 2026-08-22 (-3) | 0 in retained report evidence |
| 2026-08-23 (-2) | 0 in retained report evidence |
| 2026-08-24 (-1) | unavailable — no retained report |
| 2026-08-25 (today) | **2 on one board in the retained 18:23-09:00 UTC slice** |

First seen: **2026-08-25 08:41:53 UTC**. Last seen: **2026-08-25 08:48:40 UTC**.

## Sample log line

```json
{"event":"batch.monitor.error","crawler_type":"rss","error":"pagination fetch failed for https://<redacted-host>/<redacted-path> after <n> attempts (status=409)","level":"error","timestamp":"2026-08-25T08:41:53Z"}
```

## Root cause hypothesis

1. `apps/crawler/src/core/monitors/rss.py` treats HTTP 409 as immediately terminal for the paginated feed. Verify whether this provider uses 409 as a transient snapshot-conflict response and, if so, add a narrowly scoped bounded retry.
2. `apps/crawler/src/core/monitors/rss.py` may be constructing a pagination request that conflicts with the provider's current feed contract. Re-probe the canonical feed and offset parameters, then update the provider preset or board configuration without weakening status validation.

## Impact

One RSS posting-set refresh failed twice and may remain stale until retry. The terminal error prevents a false-empty inventory and unsafe gone detection; no broader RSS outage or worker instability is verified.

viktor-shcherb:
The same NW board recurred at 2026-08-25 16:43:33 UTC as a terminal `PaginationFetchError`, now after three connection attempts. Read-only live checks identify the underlying provider migration rather than a retryable RSS snapshot conflict:

[...]

PR #7998, opened 2026-08-25
Fence CSV sync behind deployed crawler runtime

### LINKED ISSUE #7996: Prevent CSV sync from publishing config ahead of crawler runtime
## Incident

The EPFL merge at `6d8b526160a7e6d4b555a017989dafc42ad834e3` changed both `apps/crawler/data/boards.csv` and DOM monitor runtime code. That push started both the full crawler deploy and the CSV sync workflow.

The CSV workflow has a different GitHub concurrency group and runs the requested CSV overlay with the image recorded in the host's current `.env`. The host mutation lock serializes the two transactions, but it does not ensure the full deploy wins. When CSV sync acquired it first, it published the new EPFL configs into Postgres/Redis while the previous long-lived workers remained active.

Observed effects included one old-code `rich_rows must be a bounded mapping` failure and first-cycle old-semantics counts. The same board retried successfully after the new image deployed.

## Required invariant

A standalone CSV sync must prove that the host's committed crawler deployment is compatible with the requested CSV snapshot before invoking `crawler sync`. A mixed code/config revision must be left to the full deploy's quiesce → migration/schema → inline sync → restart transaction. Do not weaken health or rollback behavior.

viktor-shcherb:
## Issues

Closes #7996
Refs #7972

## Root cause

Full deploys and standalone CSV syncs serialized host mutation but did not publish one durable runtime-plus-data generation. After runtime R0 with embedded config A, a successful data-only sync B left rollback evidence at A. A later failed R1 therefore restored A and could lose B indefinitely. The mutable shared CSV overlay also retained deleted files and could be overwritten before compatibility and lock checks.

The NW board also needed an atomic provider cutover, and EPFL needed bounded incident cleanup.

## Fix

- Publish release format v3 as one authoritative generation binding hashed Compose, environment, success marker, immutable image identity, runtime contract, exact CSV tree, canonical per-file SHA-256 manifest, Git data contract, and data revision.
- Build revision- and run-scoped immutable CSV archives, pass the Git data contract and archive digest to the host, reject tamper/residue/deletion mismatches, mount the verified tree read-only for sync, and atomically promote only after success.
- Source standalone sync credentials and executor image from the verified committed generation; reject mutable live-environment drift.
- Journal the sync-to-pointer seam durably. Failed candidates restore the prior committed snapshot; post-sync promotion crashes finish promotion; later sync/deploy work resolves the journal before mutation.
- Before the first v3 deploy, re-sync and promote the exact pre-push main CSV tree with the committed old runtime. A deploy cannot arm rollback until this succeeds.
- Make full-deploy rollback mount the immediately previous committed CSV tree read-only into the previous image. Legacy embedded CSV fallback is not allowed.
- Preserve v3 data evidence through independent Murmur generation publication.
- Classify all four process-cached runtime taxonomies plus the runtime-contract derivation script for crawler image and VERSION gates.
- Keep the runtime/data ordering fence, explicit main ref refresh, EPFL cleanup, and atomic NW Teamtailor to WTTJ repair from the prior head.
- Keep crawler VERSION at 0.13.520.

## Verification

- Full crawler suite: 9,390 passed, 68 skipped.
- Focused deploy and publication tests: 48 passed, including A to data-only B to failed R1 rollback, failed candidate recovery, exact-tree deletion/residue/tamper rejection, archive digest mismatch, live-env drift, and bootstrap ordering.
- Workflow, runtime-contract, and version tests: 72 passed.
- Ruff, Ruff format, Pyright, Actionlint, ShellCheck, Bash syntax, and diff checks passed.
- PostgreSQL NW E2E remains included in CI; local Docker is unavailable.
- Exact head: 43ed9ff3bf195a536f791db3fabd8efc24831862.


viktor-shcherb:
Exact-head adversarial review of `43ed9ff3bf195a536f791db3fabd8efc24831862`: **changes requested; keep draft**.

1. **P1 — a delayed/failed data publication can deadlock every later runtime deploy.** In `--bootstrap-current`, an already-v3 active release exits only when its data contract equals `JOBSEEK_PREVIOUS_DATA_CONTRACT_SHA256`; any mismatch is fatal. But the sync workflow re-derives the required runtime contract from *current* main immediately before SSH. Concrete sequence: active `R0/A`; merge data-only `D(B)` but its sync is delayed; merge runtime-only `R1`; the D workflow now waits for committed R1, while the R1 deploy sees the valid active v3 tree A != pre-push main B and aborts in bootstrap. Neither side can advance. The same circular wait follows a failed C publication that correctly leaves B active. For an existing valid v3 generation, rollback evidence should remain the exact actually-active tree (B/A), not require it to equal pre-push Git main; the Git candidate is needed for legacy -> v3 bootstrap. Add an executable regression for this interleaving, not only the happy legacy bootstrap ordering assertion.

2. **P2 — the new immutable artifacts have no retention or cleanup and will exhaust the host.** Every sync leaves both `/home/deploy/csv-candidates/<revision-run-attempt>/csv-snapshot.tar` and a ~7.4 MB `.crawler-release-generations/data-*` copy forever; failed/incomplete candidates and generations are also retained. At today's tree size that is about 15 MB per CSV publication (roughly 10 GB after ~670 publications), while deploy disk pruning only touches Docker state. Delete consumed candidate archives and add safe bounded pruning that always preserves the active generation, journal-referenced previous/target generations, and the intended delayed-rollback window; cover failure/SIGKILL residue too.

I also verified the assigned remote head exactly, reviewed the full diff and transaction states, and ran `bash -n`, 72 Node workflow/runtime/version tests, and the 52 focused deploy/migration pytest tests successfully. All 32 GitHub CI/CodeQL checks are green, but they do not cover the circular-wait or retention cases above.

viktor-shcherb:
Exact-head independent re-review of `34094580589b2fcd329f68e570cda1c1983170af`: **changes requested; keep draft**.

The two blockers from my `43ed9ff3` review are materially fixed: verified v3 active data is now authoritative across a later deploy, the delayed `D(B)` path checks the already-active data contract before its older runtime contract, and consumed/stale artifact pruning has been added under the mutation lock. The executable `R0/A -> D(B) -> R1` / failed-C-leaves-B test passes. I also re-verified the exact remote head, ran shell syntax checks, all 72 Node workflow/runtime/version tests, and 55 focused deploy/migration pytest tests; all 32 GitHub CI/CodeQL checks are green.

Two fail-closed gaps remain:

1. **P1 — a Git data contract is still not bound to the CSV bytes verified on the host.** `prepare_candidate_generation` verifies the archive SHA and exact per-file manifest, but then writes the independently supplied `DATA_CONTRACT_SHA256` verbatim into `release.manifest`. The full deploy likewise extracts `/app/data`, computes its manifest, and records an independently supplied Git contract without comparing them. Concrete failure: candidate/image tree A is paired with contract B, sync applies A, and the generation is promoted claiming B; a later B publisher hits `active_data_contract_matches` before runtime/archive verification and reports success as a no-op, leaving A live. Hash-shaped SSH inputs and a transfer digest do not prove this semantic relation. Make the no-op identity host-recomputable from the exact canonical manifest/tree (or pass and verify a canonical expected manifest digest for data-only, legacy bootstrap, and full-image extraction), and add an adversarial A-bytes/B-contract rejection test.

2. **P2 — the retained “newest safe generations” window is not actually restricted to verified generations.** `is_completed_generation` checks only that filenames exist, the format token is 1/2/3, and (for v3) a data directory/manifest exist; it never verifies the recorded digests or exact CSV tree. Therefore five newer crash-corrupted or structurally complete failed residues can occupy the keep window and, after the grace period, cause a valid older rollback generation to be deleted. The current retention test even models its newest entries with placeholder, internally inconsistent evidence. Reuse the full generation verifier (without mutable globals) when selecting the safe window and test corrupt newer entries against a valid older rollback. Also include stale `.crawler-forward-data-*` staging directories in crash cleanup: SIGKILL bypasses both current cleanup calls and no later path prunes them.

viktor-shcherb:
Exact-head independent adversarial review of `a6f87774429419e6825e0358f9026446d5fa203a`: **changes requested; keep draft**.

The prior data-contract binding, large-blob hashing, delayed `D(B)` / failed-C sequencing, verified-v3 retention, stale candidate/forward-staging cleanup, committed env/image sourcing, legacy bootstrap, and taxonomy/derive version-classification blockers are materially resolved. Three release blockers remain:

1. **P1 — the new pre-deploy rollback archive cannot be built in the real deploy job.** The `deploy` job's checkout at `.github/workflows/deploy-crawler-browser.yml:278` still uses `actions/checkout`'s default `fetch-depth: 1`, but `Build exact pre-deploy CSV rollback candidate` later executes `git archive "$PREVIOUS_REVISION"`. In a depth-one checkout, `github.event.before` is not present. I reproduced the exact topology by fetching only this PR head: `cat-file` could not resolve base `eb0c96060a59b077bf7220090e311d1b010ead5f`, and `git archive` exited 128 with `fatal: not a tree object`. The first post-merge crawler deploy will fail before SSH. Fetch full history or explicitly fetch and validate the exact `before` commit, and add a shallow-checkout regression.

2. **P1 — a failure after the new release pointer is published can restore the old runtime with the new CSV tree.** `publish_active_deploy_release` calls `verify_active_deploy_snapshot` in the parent shell, so `ACTIVE_RELEASE_FORMAT`, `ACTIVE_DATA_SNAPSHOT`, and `ACTIVE_DATA_FILES_MANIFEST` now name the new generation D. Rollback activates the saved old generation B, but re-verifies it only inside a subshell at `deploy.sh:1077`; those assignments do not propagate. `rollback_sync_previous_config` then uses the still-parent-held D globals at `deploy.sh:958-960`, syncing D with the restored old image/env before restarting it. Any error while rollback remains armed after pointer publication (the committed shim check, pruning, or a publication-tail I/O failure) takes this path. Rehydrate and verify the restored generation in the parent shell before config restore, and add an executable post-pointer-failure regression proving B—not D—is mounted and re-synced.

3. **P1 — the Murmur carry-forward path does not enforce the v3 fail-closed contract used by the other publishers.** Its `verify_exact_csv_tree` does not require the data root or manifest to be regular non-symlinks, so it can follow and then `cp -a` unsafe external evidence. For `HAS_IMAGE_OVERRIDE=0`, `verify_release_generation` does not reject an existing override, and the later code ignores the manifest flag and applies any `rollback-images.override.yml` merely because the file exists (`deploy-murmur-shim.yml:565-568`), even though it has no attested digest. It also keeps the runtime-contract pair optional for format 3, while the CSV host verifier requires it. Make Murmur's v3 verification equivalent to the host verifier: regular non-symlink data/manifest, exact tree/contract, mandatory matching runtime evidence, and override presence exactly controlled by `HAS_IMAGE_OVERRIDE`; cover symlinked data/manifest, stray override with flag 0, and missing runtime evidence.

Evidence at this exact remote head: `bash -n` and ShellCheck are clean; all 74 Node workflow/runtime/version tests pass; all 53 focused deploy-script pytest tests pass; Required CI and every CI/CodeQL check are green. Those checks do not exercise the depth-one push deployment, the post-publication rollback state leak, or the Murmur v3 adversarial cases above.

viktor-shcherb:
Exact-head independent re-review of `a9099eba0ba391b25c26245b0def02762c4a4c96`: **changes requested; keep draft**.

The three blockers from my `a6f8777` review are materially fixed: the deploy checkout can now archive the exact ancestry-validated pre-push revision, post-pointer rollback rehydrates the old generation in the parent shell before restoring its config, and Murmur v3 now enforces the same tree/runtime/override evidence. Shell syntax and ShellCheck are clean; all 74 focused Node tests and all 55 deploy-script pytest tests pass; Required CI and all CI/CodeQL checks are green.

One first-rollout blocker remains:

**P1 — a real legacy format-1/2 release cannot be bootstrapped to v3.** `prepare_candidate_generation` copies the active legacy `environment.env` and `success.env` unchanged, writes `RELEASE_FORMAT_VERSION=3`, then calls `verify_release_generation`. The v3 verifier now correctly requires exactly one matching `JOBSEEK_RUNTIME_CONTRACT_SHA256` in both files. However, neither the format-1 release writer on the PR base nor Murmur's format-2 legacy bootstrap records that field, so the first full deployment reaches this path with no runtime evidence and aborts before it can publish rollback state. I reproduced the actual shape with a valid format-2 active generation lacking the field: `--bootstrap-current` returned 1, left `legacy.test` active, and stranded the newly prepared `data-<revision>.*` generation.

Add a trustworthy migration that attests the **previously active runtime** when converting legacy evidence to v3; do not stamp the incoming runtime contract onto the old image/config, because that would make rollback evidence lie about compatibility. The acceptance test should execute at least the production-relevant format-2-without-runtime → `--bootstrap-current` → verified v3 path and assert the resulting old-runtime contract, exact pre-push CSV tree, and active pointer. A first-full-deploy regression from that legacy state would cover the integration boundary as well.

viktor-shcherb:
Exact-head independent re-review of `7aa99d9923dd3c1fc4f097b4f2d79e121ff7c449`: **changes requested; keep draft**.

The previous legacy-bootstrap blocker is materially fixed for the production-shaped format-2 path: the candidate derives the old runtime from immutable `github.event.before`, carries its contiguous first-parent runtime epoch, checks active revision and env/success image identities, validates pinned Compose images plus the format-2 override, durably adds the old runtime pair, and the executable format2 → bridged-v3 → failed-new-deploy case restores old runtime/data. The earlier data-contract, rollback-parent, archive-history, Murmur v3, retention/journal, delayed-publication, and large-blob fixes remain intact. Shell syntax and ShellCheck are clean; all 75 focused Node tests and 57 deploy-script pytest tests pass; Required CI and every CI/CodeQL check are green.

Two fail-closed gaps remain:

1. **P1 — format 1 can promote an unattested image override into the v3 rollback contract.** `verify_legacy_runtime_evidence` validates override presence/digest only for source format 2, but `prepare_candidate_generation` copies `rollback-images.override.yml` whenever the file exists, independent of the verified source format, and then records it as `HAS_IMAGE_OVERRIDE=1`. I reproduced a valid format-1 generation containing an otherwise-unattested override that replaced `worker-1` with a different digest-pinned crawler image: `--bootstrap-current` returned 0, synced/promoted the bridge, and newly blessed the override digest in its v3 manifest. The later deploy verifier checks that resolved images are pinned but does not bind service images to the env/success identities, so a wrong-but-pinned override can become rollback runtime; a mutable one instead strands the first rollout only after the pointer/data mutation. Format 1 must reject any override, including dangling symlinks, before generation creation/sync; format 2 should copy only the exact verifier-derived override. Add an executable wrong-pinned format-1 override regression asserting no sync, pointer change, or generation residue.

2. **P2 — the bridge provenance is not durable across its other consumers.** The CSV host verifier strictly reconstructs the original env/success digests and checks the attested source revision, but `apps/crawler/deploy.sh` and the Murmur `verify_release_generation` path ignore `LEGACY_*` and `runtime-attestation.env` entirely. I altered a bridged attestation so its recorded source revision was outside the compatible epoch and updated its attestation digest: the strict host verifier returned 1, while both deploy and Murmur verifiers returned 0. Murmur's format-3 carry-forward also copies data/runtime but drops the attestation file and every bridge field, permanently converting the release to generic v3 without first validating that provenance. This matters after a failed first full deploy leaves the bridge active: an independent Murmur publication can erase the only durable proof tying the injected runtime contract to the real legacy release before the next retry/rollback. Make every bridge consumer validate the same strict metadata and have Murmur preserve/re-hash it (or record an equivalently strict transitive conversion). Add executable corrupt-bridge rejection tests for deploy and Murmur plus bridge → Murmur carry-forward → failed crawler deploy rollback coverage.

viktor-shcherb:
Exact-head independent re-review of `12b0848017a082371355e86b742ffc745ae0ca2b`: **changes requested; keep draft**.

The two findings from my `7aa99d9` review are materially fixed: format 1 now rejects regular and dangling override residue before sync/pointer mutation; format 2 carries only its verifier-derived exact override; and the dependency-free bridge validator is delivered to the CSV host, crawler deploy, and Murmur paths. It reconstructs the original source evidence, all three consumers reject corrupt bridge provenance, Murmur preserves it transitively, and the new bridge → Murmur → failed crawler deploy regression restores the old runtime/data.

One release blocker remains:

**P1 — the first rollout requires the new verifier to exist in the old active deploy directory before it installs it.** `scripts/verify-crawler-release-bridge.py` is new relative to the exact base, but it was added to `DEPLOY_SPEC_FILES`. After the legacy bootstrap has already synced and promoted the v3 bridge, `snapshot_active_deploy_specs` loops over that list and requires `/home/deploy/scripts/verify-crawler-release-bridge.py`. The crawler job copied it only to `/home/deploy/incoming/scripts`; Murmur copied it to `incoming-shim`, and data sync uses `csv-sync-runner`. `activate_staged_deploy_specs`, which would create the active path, runs only *after* the snapshot. Therefore the first deployment fails with “active deploy artifact is unavailable” after bridge publication but before rollback is armed. Every retry sees the valid v3 bridge, takes the bootstrap no-op, and fails at the same missing file, so the crawler cannot advance without manual host mutation.

Handle a newly introduced deploy artifact explicitly at the old-spec snapshot boundary without weakening rollback: the rollback archive must faithfully represent the pre-rollout artifact set, while the staged verifier must become active only at a transaction point whose failure can restore/remove it correctly. Add an executable first-rollout test seeded with the exact pre-PR `DEPLOY_SPEC_FILES` (no active verifier) that reaches staged activation, plus a forced later failure proving rollback restores the old artifact set and a retry succeeds.

Evidence at this exact head: `bash -n`, ShellCheck, and Python compilation are clean; all 75 focused Node tests and 59 deploy-script pytest tests pass. Hosted CI/CodeQL was still running Test Web, Test Crawler, Web Typesense E2E, and JavaScript CodeQL when this blocker was posted; every completed check was green.

viktor-shcherb:
Exact-head independent re-review of `784a1e87b51c6d6322d7f7177e5c1b4eb652b94e`: **APPROVED** — no blocking findings.

The prior first-rollout blocker is materially fixed. The deploy now snapshots the exact pre-rollout artifact set as ordered, hash/mode-bound `PRESENT`/`ABSENT` evidence before arming rollback or activating staged specs. Rollback restores verified present artifacts and removes previously absent ones only after the old release/config has been re-verified, so partial activation and later failures faithfully recover the legacy set. The staged bridge verifier is used for the initial legacy verification, becomes active inside the rollback-protected transaction, and a subsequent retry can complete without a loop.

Evidence at this exact frozen head/base (`784a1e87` / `74d36782`): shell syntax, ShellCheck, Python compilation, and `git diff --check` are clean; all 75 focused Node workflow/runtime/version tests and all 60 deploy-script pytest tests pass. I also ran the new first-rollout rollback/retry regression three additional consecutive times successfully. The remote PR head and `main` still match the reviewed SHAs, and Required CI plus every reported CI/CodeQL check are green.

viktor-shcherb:
Post-merge deployment verification complete. The exact merged revision `5ed23e2f1a10806517540c790bf75210c53768ce` deployed successfully in Hetzner run https://github.com/colophon-group/jobseek/actions/runs/32903244348 (attempt 2; attempt 1 was an external GitHub action-download DNS failure). Every crawler service reports version 0.13.520. Public production inventory verification: EPFL has 121 active jobs / 121 distinct canonical URLs, including HQC 2 and LFIM 3; NW has 9 active jobs / 9 distinct canonical URLs, all Welcome to the Jungle, with 0 legacy NW URLs. Issues #7996 and #7972 are closed, and duplicate PR #8002 was closed as superseded.

PR #8049, opened 2026-08-26
Implement checked record framing

viktor-shcherb:
## Summary

- add candidate-only Python and Go canonical uint64 varint record framing
- enforce prefix-inclusive checked caps before hostile arithmetic, slicing, conversion, or allocation
- keep Python concatenated decoding linear with a no-copy memoryview remainder and exact trailing rejection before payload copying
- classify Go unexpected EOF after a record begins as ambiguous EOF while accepting completed records
- share deterministic boundary, malformed-prefix, giant-length, 16 MiB trailing, fragmentation, and backpressure vectors

## Verification

- 31 focused Python contract/runtime tests passed
- 9,577 full crawler tests passed; 69 skipped
- full lane Pyright: 0 errors
- Ruff check and format passed
- Go race tests repeated three times and Go vet passed
- deterministic vector generation and SHA-256 manifest passed
- pre-commit and pre-push hooks passed

## Coordination

This is the authorized clean replacement for #8048, which was closed unmerged after a single-writer collision. It is built as one commit directly from reserved main `a380a563d9d7090b76e1505448082137fa267213`; no commit was cherry-picked from the retired branch. The framing package remains candidate-only and does not activate or package a runtime contract.

Reservation: https://github.com/colophon-group/jobseek/issues/7937#issuecomment-5425538899

Closes #8041
Parent #7937

PR #8051, opened 2026-08-26
Stabilize BrassRing pagination snapshots

### LINKED ISSUE #7477: [daily-error-review] browser.brassring_incomplete_inventory (novel 13; two boards)
## Summary

Browser-1 emitted **13** terminal BrassRing monitor failures because the provider's advertised row total exceeded the number of valid unique jobs assembled after pagination. The failures occurred during the explicit **2026-08-15 09:01 UTC -> 2026-08-16 09:01 UTC** review window and affected two boards. Application logs retain only the final post-deployment generation from about 00:41 UTC onward, so 13 is an observed-slice count rather than a supported full-window trend. This provider-specific class first became possible after the BrassRing monitor from PR #7230 deployed in the review window; the existing Workday incomplete-inventory issue #7244 is a different service and exception class.

## Signal

The unit is a terminal browser-1 `batch.monitor.error` with `crawler_type=brassring` and stable message `BrassRing returned <valid> valid unique jobs for <expected> expected rows`.

| day | count |
|---|---:|
| 2026-08-12 (-4) | unavailable — no retained report |
| 2026-08-13 (-3) | 0; class not deployed |
| 2026-08-14 (-2) | 0; class not deployed |
| 2026-08-15 (-1) | 0; class not deployed in the report window |
| 2026-08-16 (today) | 13 across two boards in the retained 00:41-09:01 UTC slice |

The failures ran from **01:18:25 UTC through 08:19:03 UTC**. Observed deficits ranged from one to five jobs. Earlier zeroes establish deployment context but are not used to claim a rate regression.

## Sample log line

```json
{"event":"batch.monitor.error","crawler_type":"brassring","error":"BrassRing returned <valid> valid unique jobs for <expected> expected rows","level":"error","timestamp":"2026-08-16T01:18:25Z"}
```

## Root cause hypothesis

1. `apps/crawler/src/core/monitors/brassring.py` compares the provider's advertised total with jobs remaining after parsing and URL deduplication, but does not retain whether the deficit came from invalid rows or duplicate canonical URLs. Track those reasons separately and validate against the provider's stable job identity while preserving the fail-closed completeness guard.
2. `apps/crawler/src/core/monitors/brassring.py` snapshots the total at the first page while the upstream inventory can change during a multi-page run. The per-page total check may still miss churn that replaces or duplicates individual rows without changing the aggregate. Add a bounded consistency/retry pass or a stable-ID inventory comparison before accepting or rejecting the run.

## Impact

Two BrassRing posting-set refreshes repeatedly failed and may remain stale until a consistent run or monitor fix. The explicit failure prevents silently accepting an incomplete inventory or tombstoning omitted postings. Browser-1 remained running and healthy; no restart, OOM, or broad browser outage was verified.

viktor-shcherb:
Closes #7477

## Problem
BrassRing defaults multi-page search results to LastUpdated ordering. Listings moving within that ordering while a crawl is in progress can repeat a requisition on adjacent pages or leave the collected row count short, so the existing completeness guard correctly fails the whole run. Production browser logs still showed terminal repeats after both retry attempts.

## Change
- switch multi-page inventories to the provider's visible Alphabetical sort before collecting the authoritative page-one snapshot
- preserve sequential UI pagination and its Angular page-state barrier
- fail closed when the stable sort is unavailable or a non-empty inventory disappears during the switch
- leave one-page boards unchanged, because they have no pagination/sort control
- bump crawler VERSION to 0.13.525

## Evidence
Exact changed code was exercised live on 2026-08-26:
- ADM US: 567 rows / 567 unique requisitions in 17.8s
- ADM Global: 316 / 316 in 13.32s
- ADM Early Careers: 4 / 4 in 5.31s (single-page path)

Concurrent and direct sequential XHR pagination were rejected during diagnosis because the provider returned duplicate requisitions; the verified UI path is deliberately retained.

## Tests
- `uv run pytest tests/test_brassring.py` (22 passed)
- `uv run ruff check src/core/monitors/brassring.py tests/test_brassring.py`
- `uv run ruff format --check src/core/monitors/brassring.py tests/test_brassring.py`
- live monitor validation listed above


viktor-shcherb:
## Independent exact-head review — PASS

Independent reviewer `/root/review_pr8051` approved exact head `ecabd01c0951c7a51ecd917361c9adbe25fbc879` against base `a380a563d9d7090b76e1505448082137fa267213` with no actionable findings. Focused tests passed 22/22, Ruff is clean, and an independent live ADM Global run returned 316/316 unique jobs in alphabetical order. Exact-head Required CI, Workflow Security, PostgreSQL/Typesense E2E, and CodeQL are green.

This approval is head-specific. Keep the PR draft: the Typesense capacity/deploy gate remains closed, and #8049 has since moved `main`; any later rebase/version-only delta needs an exact bounded delta recheck before merge.

PR #8058, opened 2026-08-26
Prepare 8 GB Typesense host memory policy

viktor-shcherb:
## Summary
- stage an 8 GB/CX33 policy with a 6 GiB hard limit, 5 GiB reservation, and 6 GiB memory-plus-swap ceiling
- fail closed before host mutation when expanded policy sees less than 7 GiB total memory
- enforce the host→backup contract in a non-runtime verifier run by the host workflow
- keep an explicit, temporary legacy tuple only across the staged transition; #8059 removes it after promotion

## Safety
- this PR does not resize Hetzner or dispatch production
- protected promotion explicitly selects `expanded`; production mutation remains manual `workflow_dispatch`
- the current 4 GB host is rejected before state directories or deployment locks are created
- the new backup installer requires the expanded tuple; the runner bridge permits the old timer to remain valid only until host cutover quiesces it
- no `apps/crawler/**` path changes, so this bootstrap PR does not trigger crawler runtime deployment
- `deployment-hold:crawler` remains active on #6632

## Verification
- 63 backup/host deployment tests passed
- non-runtime cross-component memory contract verifier passed
- expanded memory-capacity and credential-conformance self-tests passed
- actionlint, shell syntax, ShellCheck, Python compilation, Ruff, and diff checks passed

Supports #6632. Follow-up: #8059. Neither issue is closed by this bootstrap PR.


viktor-shcherb:
Independent exact-head review of `f612fde12d94ac16fbd544ba3197595de547133b` found two P1 blockers: (1) the installed `scripts/jobseek-data-backup.py` still rejected the expanded tuple, so the mandatory fresh backup smoke would fail; (2) the focused crawler-test delta is runtime-classified, producing the observed Version bump/Required CI failure and making the bootstrap PR unmergeable under #6632. One bounded repair pass is now reserved: restore the crawler test byte-for-byte, update the actual backup execution contract, add equivalent cross-component assertions under `deploy/typesense-host/` and run them from the host validation workflow, then request finding-only recheck on the new exact head. No adjacent redesign or second repair loop.

viktor-shcherb:
Sole bounded repair pushed at exact head `c35b849aa59d267797e589376de71a9436a9f15a`: the runtime-classified crawler test delta is removed byte-for-byte; equivalent ordering and full host→backup→runner tuple checks now live in executable `deploy/typesense-host/verify-memory-contract.py` and run in the host workflow. The actual backup runner now recognizes the expanded tuple during an explicit two-policy migration bridge, while the new backup installer requires expanded; #8059 removes legacy recognition after cutover. Focused backup/host suite: 63 passed. Requesting finding-only recheck of the two prior P1s on this exact head; no adjacent review.

viktor-shcherb:
Independent finding-only recheck: **APPROVE** exact head `c35b849aa59d267797e589376de71a9436a9f15a` against exact base `b01982af87348a330b4ea91bb6f982c21127535f`. Both prior P1s are resolved: the installed runner accepts expanded 6/5/6 through the mandatory fail-closed smoke with #8059 bounding legacy removal, and the final diff has no crawler-runtime path while the non-runtime cross-component verifier is invoked and green. No direct repair regression found.

Issue #7936, opened 2026-08-25
[pilot] Compare Python with a bounded Go slice


viktor-shcherb:
> [!IMPORTANT]
> **Superseded scope (2026-09-07):** The original body below is historical and must not be used as current acceptance criteria or dependency ordering. The authoritative pilot-first reset is https://github.com/colophon-group/jobseek/issues/7935#issuecomment-5569093871, followed by the issue's latest pilot-first rescope comment. Broad migration, exact evidence, projected-load, zero-Chromium, and production authority are not implied.

Parent: #7935

## Outcome

Estimate what it would cost to operate the crawler at the same projected load
with:

1. the current Python + Playwright/Chromium implementation; and
2. the replacement Go + self-hosted Lightpanda implementation.

This is an apples-to-apples crawler-runtime capacity and ROI comparison. It is
not a total-platform cost model. CHF 50/month is the current allocated budget
and is already known to be insufficient for today's crawler load; treat it as
an underfunded reference whose shortfall must be quantified, never as a cost
baseline or acceptance ceiling.

## Projected crawler load

Use the same frozen workload for both implementations:

- 10 million configured boards;
- 100 million active postings;
- at least 1 million successful monitor cycles due in one hour;
- at least 5 million successful detail cycles due in one hour during catch-up;
- the same measured browser share, response-size distribution, retry rate,
  provider mix, schedule shape, and origin-policy constraints;
- a bounded 2x arrival/recovery scenario and one crawler worker/browser
  instance loss to expose operational headroom.

If later production evidence changes an input, publish a new dated workload
revision and apply it identically to Python and Go. Never change the workload
or silently drop board/capability classes to improve one implementation.

## Crawler-only cost boundary

Include every runtime resource required to execute that projected crawl load:

- Python or Go monitor/scrape workers;
- Playwright/Chromium or self-hosted Lightpanda browser execution;
- crawler scheduling/queue/Redis capacity attributable to execution;
- crawler-side drain/export/sync process compute when those processes are part
  of the implementation being compared;
- proxy and network cost attributable to crawler origin traffic;
- service redundancy and recovery headroom required for the crawler runtime.

Exclude the products the crawler writes to or supports:

- Postgres storage/service cost;
- Typesense nodes, indexes, and search-query capacity;
- R2 object storage and operation charges;
- web hosting, application traffic, backups, and unrelated telemetry or
  control-plane services.

Include only a separately reported attributable delta if the implementation
[...]

PR #8180, opened 2026-08-29
Fix installed crawler build version

### LINKED ISSUE #8179: [crawler] Make installed build_info version authoritative
Parent: #7936

## Blocker

The automatic 0.13.592 deployment succeeded and all six long-lived crawler targets are healthy with the new process-tree metric family, but every target exports `crawler_build_info{version="unknown"}`. The 86,400-second projected-load capture must not start without an exact release identity.

## Cause

The installed wheel imports `src.metrics` from site-packages, while the runtime resolver probes for a checkout-relative `VERSION` beside the installed module. The wheel metadata already carries the derived crawler version; the lookup does not use it. The baked `/app/src` and `/app/VERSION` paths must not be treated as installed-package authority.

## Scope

- Reserve crawler VERSION 0.13.593.
- Use installed distribution metadata for installed wheel/image runtime version identity.
- Preserve structurally verified checkout `VERSION` behavior for development, avoiding stale editable-install metadata.
- Add isolated installed-wheel and slim-image release assertions covering worker-1/2/3, browser-1, exporter, and drain entry paths without network or database side effects.
- Make `unknown` fail release acceptance.

## Out of scope

Process-tree semantics, deployment workflow, infrastructure, Lightpanda behavior, pricing, ROI calculations, and any manual production mutation.

## Acceptance

- Checkout tests read the checked-out VERSION even if installed metadata is stale.
- Installed wheel/image resolve exact 0.13.593 from distribution metadata and never from baked checkout-looking files.
- All six long-lived role entry paths initialize/export exact 0.13.593 build info before DB/HTTP initialization.
- Full local and hosted gates pass; draft PR stops for independent review before merge.

viktor-shcherb:
Closes #8179
Parent: #7936

## Blocker

The automatic 0.13.592 deployment succeeded, but all six long-lived crawler targets exported crawler_build_info with the fallback version unknown. The projected-load capture cannot start without an exact release identity.

## Change

- Resolve installed runtime identity from the jobseek-crawler distribution metadata carried by the wheel.
- Keep structurally verified checkout/dev imports on the checked-out VERSION file so stale editable metadata cannot override source state.
- Reject missing, blank, unknown, malformed, or unbounded installed identities before build-info labeling, sampler/listener startup, database pools, or HTTP clients.
- Preserve the repository's bounded derived build form: major.minor.patch+build.commit-count.gsha.
- Add one release probe used by isolated-wheel and slim-image CI. It exercises worker-1, worker-2, worker-3, browser-1, exporter, and drain through the real metrics initializer, then stops before DB/HTTP side effects.
- Bump crawler VERSION to 0.13.593.

## Scope boundaries

No process-tree metric semantics, deploy workflow, infrastructure, Lightpanda behavior, pricing, ROI, or crawler workload changes.

## Local evidence

- Focused build-info/entry tests: 23 passed.
- Full crawler suite with declared enrich extra: 10,929 passed, 82 skipped.
- Ruff check and format: passed.
- Pyright: 0 errors.
- Actionlint: passed.
- Repository Node policy tests: 153 passed.
- Isolated built wheel installed into a clean venv: exact 0.13.593 and all six role probes passed from site-packages.
- Local Docker daemon was unavailable; the hosted linux/amd64 slim-image assertion is the authoritative remaining image gate.

## Capture boundary

No capture window starts from this PR. After reviewed merge and a successful automatic deployment, all six targets must expose exact 0.13.593 and complete metrics before the exact 86,400-second window is marked. Cumulative failure, gap, and restart counters are snapshotted at start and must have zero delta during the window; lifetime values need not be absolute zero.


viktor-shcherb (commented):
Formal exact-head review of `0e7a95183a28623934bb3fb5e2de309761eb4af2`: **1 blocker**.

**Blocker — source/editable precedence is structural, not provenance-verified.** In `src/metrics.py`, `_read_version()` trusts the source `VERSION` whenever `is_source_checkout()` is true. That predicate only checks for a checkout-shaped tree (`pyproject.toml` plus `src/shared/constants.py`); it does not establish that the `jobseek-crawler` distribution is absent or that its PEP 610 `direct_url.json` identifies this exact checkout as editable. The production image deliberately contains `/app/src`, `/app/pyproject.toml`, and `/app/VERSION`. If import resolution selects that copied tree, a noneditable installed runtime will accept the copied `VERSION` (or `unknown`) and ignore installed distribution metadata, contrary to the fail-closed authority contract. The unit tests monkeypatch `is_source_checkout()`, and the image probe uses isolated installed-package resolution, so neither exercises this hostile-but-real ambiguity.

Bounded repair: resolve `jobseek-crawler` distribution provenance centrally. Permit source `VERSION` only when no installed distribution exists, or when matching PEP 610 editable metadata points to the same checkout; installed noneditable metadata must win despite checkout-looking files, and mismatched/ambiguous provenance must fail closed. Add tests for (1) matching editable metadata with stale installed version uses source, (2) noneditable metadata plus misleading source paths still uses metadata, and (3) mismatched editable target fails; extend the slim-image gate so misleading `/app/VERSION` or source-first import resolution cannot override metadata.

Everything else checked for this head is aligned: the distribution key is `jobseek-crawler`; accepted release strings are bounded; missing/blank/`unknown`/malformed installed values fail before sampler, listener, DB, or HTTP; source, wheel, and runtime agree on `0.13.593`; clean-wheel arbitrary-CWD and actual Linux/amd64 slim six-role probes are green; workflow/security gates are green; scope is limited to this contract with no process-tree, deploy/rollback, or ROI changes. The draft-only deploy-gate failure is expected and grants no merge/deploy authority. General Test Crawler/Test Web jobs were still running when this review was posted.

viktor-shcherb (commented):
Formal exact-head re-review of `0520f321479847ec11e269b2fa989d1f1873b632`, limited to the prior provenance blocker: **no blockers found**.

The repair is a single direct descendant of `0e7a95183a28623934bb3fb5e2de309761eb4af2` and implements the required authority order for the `jobseek-crawler` distribution. A structurally verified checkout may read source `VERSION` when the distribution is absent; when a distribution exists, source may override its version only when valid PEP 610 `direct_url.json` marks an editable `file:` target whose resolved absolute path exactly matches the imported checkout. Matching editable provenance still requires the structural checkout predicate and a valid bounded source version.

The copied-image ambiguity is closed: absent/noneditable direct-URL provenance keeps installed metadata authoritative even when `/app/src`, `/app/pyproject.toml`, and a misleading `/app/VERSION` make the import look like a checkout. Malformed JSON/shape, mismatched editable targets, non-file or relative editable URLs, and query/fragment-bearing editable URLs fail closed. Installed blank/`unknown`/malformed identity is rejected by `_read_version()` before the build-info label, process-tree sampler, metrics listener, database, or HTTP client can initialize.

Coverage for this boundary is 37 parametrized cases: 26 build-info/version cases plus 11 metrics-server/role cases, including matching editable source precedence, missing and noneditable provenance, misleading source shape, mismatch, malformed/relative/query-bearing provenance, invalid installed versions, and four CLI command paths. Hosted `Test Crawler` is green.

Release and runtime evidence also agrees on `0.13.593`: the clean wheel installs `jobseek-crawler==0.13.593` and verifies all six long-lived roles; installed image parity verifies both isolated site-packages resolution and source-first `PYTHONPATH=/app` with `/app/VERSION=0.13.111`, each yielding exactly `crawler_build_info{version="0.13.593"} 1` for six roles with DB/HTTP/listener/sampler effects forbidden. The actual Linux/amd64 slim build probe is green. Required CI, CodeQL, dependency review, and workflow security are green.

The repair commit changes only the provenance resolver, its focused tests/probe, and the two existing CI gates. The full PR remains scoped to build-version observability; no process-tree behavior, deployment/rollback semantics, or ROI work changed. The PR remains draft; the draft-only deploy-gate failure is expected and grants no approval, readiness, merge, or deploy authority.

PR #8280, opened 2026-08-30
Cache public API responses behind edge rate limiting

### LINKED ISSUE #8261: [Public API] Make anonymous search responses CDN-cacheable without weakening rate limits
## Problem

The rolling production 12-hour view on 2026-08-30 showed `/api/v1/search` at approximately **49 invocations / 5s Active CPU**. Every origin execution performs an Upstash per-IP rate-limit operation before Typesense work.

Responses advertise shared-cache freshness, but client-specific `X-RateLimit-*` headers and origin-side enforcement make shared CDN reuse ambiguous: a cache hit would bypass the limiter, while caching one caller's remaining/reset values would leak incorrect policy state to another caller.

## Scope

- Measure current edge cache results for representative identical search queries.
- Define a rate-limit architecture in which shared CDN hits cannot weaken abuse protection.
- Prefer existing Vercel Firewall enforcement for edge-wide controls; retain origin limits where they protect non-cacheable/mutating operations.
- Remove client-specific headers from any response made shareable, or make such responses explicitly private.
- Preserve CORS, validation, anonymous API contracts, hosted-MCP attribution, metrics semantics, and 429 behavior.
- Roll out log-first and document cache-hit versus origin-execution accounting.

## Acceptance criteria

- [ ] Repeated identical anonymous GET searches can be served from CDN without executing the Function.
- [ ] Abuse limits apply before both CDN and origin in a documented, tested way.
- [ ] No caller receives another caller's remaining/reset values.
- [ ] Validation errors and 429s have explicit safe cache policies.
- [ ] Hosted MCP and external attribution remains correct for origin executions.
- [ ] Functional, rate-limit, cache-header, and production smoke tests pass.
- [ ] A post-deploy comparison records cache results, invocations, CPU, and Upstash command change.

## Safety

Do not publish a Firewall rule change without reviewing the exact draft and preserving all existing deny/rate-limit behavior.

viktor-shcherb:
Closes #8261.

## Problem

Production already caches successful `/api/v1/*` GETs, but the shared response also contains caller-specific `X-RateLimit-*` values. A repeated `/api/v1/search` probe returned MISS then HIT with the first caller's stale remaining/reset values. The HIT also bypassed the origin Upstash limiter with no pre-cache public-API control.

## Change

- split private/non-shareable JSON responses from deterministic shared successes
- keep validation, provider, degraded, not-found, and origin 429 responses `no-store`, with CORS and caller rate-limit metadata where available
- opt public REST routes into strict Typesense handling so outage-shaped empty/degraded results cannot enter shared cache
- cache successful anonymous GETs only at Vercel using `Vercel-CDN-Cache-Control`
- make browsers and downstream intermediaries revalidate and omit `X-RateLimit-*` from shared responses
- apply the helper to search, job, companies, resolve, taxonomies, watchlists, and watchlist/create
- retain origin Upstash 30/min sliding enforcement and origin-only REST/MCP attribution
- publish OpenAPI v1.2 with the two rate-limit layers, cache-safe success contract, provider failures, and pre-cache 403 on all seven GET operations
- synchronize the MCP/public API contract version and document the edge/origin/cache contract and rollback procedure

## Unpublished Firewall draft

Vercel Hobby permits three custom rules and one rate-limit rule, so a fourth rule was rejected without staging. The reviewed draft instead modifies only `rule_deny_public_read_server_action_bursts_SZRRK0`:

- preserve position 2, enabled state, fixed-window algorithm, IP key, 60/60 threshold, and exceeded action `deny`
- preserve the existing two Server Action OR groups
- add one OR group requiring production + host `jseek.co` + GET + path regex `^/api/v1(?:/.*)?$`
- rename it to `Deny public read and API bursts` and update its description
- leave the first deny rule, trailing API/MCP log rule, and both hostname-scoped IP blocks unchanged

The Firewall change is staged but deliberately not published. It must be published by the operator before this PR merges.

## Validation

- 85 focused route/service/OpenAPI regression tests
- full web suite: 271 files / 2,121 tests
- web and MCP TypeScript
- focused ESLint and commit-hook ESLint
- MCP package build and contract verification
- production Next.js build
- bundle budget: 670.0 KB gzip
- `git diff --check`
- all required GitHub CI checks green

## Independent review

The critic found two issues in the first pass: degraded provider results could still be cached, and the Firewall log-rule documentation misattributed earlier denials. Both were fixed. Exact-head re-review at `e3721d69b5081d3d90aa585fa694126d48d3c3d2` is clean with no P0/P1/P2 findings.

### DIFF
--- a/apps/web/app/api/v1/_shared.test.ts
+++ b/apps/web/app/api/v1/_shared.test.ts
@@ -119,11 +119,43 @@ describe("checkRateLimit — Redis bypass observability (#3175)", () => {
 });
 
 describe("apiResponse status contract (#3213)", () => {
-  it("uses 200 by default but honors explicit non-2xx status codes", async () => {
+  it("uses 200 by default, honors status codes, and is never cacheable", async () => {
     const { apiResponse } = await import("./_shared");
 
+    const response = apiResponse(
+      { error: "Bad request" },
+      {
+        rateLimit: { limit: 30, remaining: 29, reset: 12345 },
+        status: 400,
+      },
+    );
+
     expect(apiResponse({ ok: true }).status).toBe(200);
-    expect(apiResponse({ error: "Bad request" }, { status: 400 }).status).toBe(400);
+    expect(response.status).toBe(400);
+    expect(response.headers.get("Cache-Control")).toBe("no-store");
+    expect(response.headers.get("Vercel-CDN-Cache-Control")).toBeNull();
+    expect(response.headers.get("X-RateLimit-Remaining")).toBe("29");
+  });
+
+  it("caches deterministic successes only at Vercel and omits caller metadata", async () => {
+    const { sharedApiResponse } = await import("./_shared");
+
+    const response = sharedApiResponse(
+      { ok: true },
+      { maxAge: 3600 },
+    );
+
+    expect(response.status).toBe(200);
+    expect(response.headers.get("Cache-Control")).toBe(
+      "public, max-age=0, must-revalidate",
+    );
+    expect(response.headers.get("Vercel-CDN-Cache-Control")).toBe(
+      "public, max-age=3600",
+    );
+    expect(response.headers.get("Access-Control-Allow-Origin")).toBe("*");
+    expect(response.headers.get("X-RateLimit-Limit")).toBeNull();
+    expect(response.headers.get("X-RateLimit-Remaining")).toBeNull();
+    expect(response.headers.get("X-RateLimit-Reset")).toBeNull();
   });
 });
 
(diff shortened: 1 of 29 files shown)


viktor-shcherb:
Independent critic re-review at exact head `e3721d69b5081d3d90aa585fa694126d48d3c3d2`: clean, with no P0/P1/P2 findings. The prior degraded-provider caching and Firewall attribution findings are fixed. The critic independently verified the six provider-backed cacheable routes, uncached `/api/v1/job` failure behavior, OpenAPI v1.2/MCP contract synchronization, 160 focused tests, web typecheck/lint, MCP verification/typecheck, and the still-unpublished one-change Firewall draft with its existing protections and IP blocks preserved.

Issue #8366, opened 2026-08-31
[watchlist cleanup] Retire legacy watchlist access and subscription restrictions


viktor-shcherb:
> **Scope reset (2026-09-07):** The authoritative intervention directives in this issue discussion supersede conflicting scope, dependency, sequencing, or pilot language below. Historical text remains for traceability. Every rendered UI/product-copy change remains blocked pending explicit linked human approval; agents may not merge.

## Outcome

Retire the legacy watchlist access and subscription model before subscriber AI-filter implementation begins:

- replace plan-specific watchlist quantities with the universal account ceiling in #8316;
- make watchlists private owner resources and retire ambient public discovery/access under #8348 while preserving a source-policy seam for future authorized sharing; and
- track #8317 as an independent notification-prerequisite workstream with its own persistence, matcher, scheduler, UI, delivery, and cost gates.

This is an orchestration issue, not authorization to bundle the three products into one implementation PR. The live child issues remain the detailed acceptance sources.

## Dynamic planning contract

This checklist and order are a current evidence-based decomposition, not a frozen issue or PR count.

- Agents may split work, add children, supersede invalid work, or reorder downstream items when reconnaissance, migration, provider, or review findings justify it.
- Any split or reordering must preserve the acceptance boundary, link replacements, and update this tracker and affected child issues before the old item closes.
- Do not broaden an active PR merely to preserve the checklist shape.
- #8317 is independent: it does not block the #8316 → #8348 cleanup path or the AI feasibility gate, and its live delivery must not be coupled to AI accepted/rejected queues.
- #8323 remains the source of truth for the AI-filter feasibility gate and downstream AI implementation. This tracker does not open model/service/UI implementation before that gate.

### Merged backend/cleanup slices

- [x] #8372 — universal 10-watchlist backend cap, destination-owner serialization, grandfathering, and copy/create-path enforcement.
- [x] #8374 — anonymous REST watchlist discovery now returns the bounded `410 Gone`; unauthenticated MCP `search_watchlists` discovery is removed.
- [x] #8375 — notification cadence/pause/enable-time persistence and durable delivery-policy ledger.
- [x] #8376 — private mutation defaults and explicit copy-source authorization separated from destination-owner capacity.
- [x] #8380 — canonical session-free watchlist matcher shared by interactive reads and future background consumers.
- [x] #8386 — providerless `off`/`shadow` scheduler core with durable claiming/window logic and no cron, Redis lock, provider callback, or live configuration.
[...]

Issue #8317, opened 2026-08-31
Add weekly watchlist notification digests


viktor-shcherb:
> **Scope reset (2026-09-07):** The authoritative intervention directives in this issue discussion supersede conflicting scope, dependency, sequencing, or pilot language below. Historical text remains for traceability. Every rendered UI/product-copy change remains blocked pending explicit linked human approval; agents may not merge.

## Summary

Add opt-in weekly email notifications for watchlists, available to all users. Send one consolidated digest per user per week, aggregating matches from every enabled watchlist.

There is no cadence selector in the first release. The implementation must keep cadence as an internal policy so daily/monthly options can be added later without rewriting matching, scheduling, or delivery state.

Related: #8316 replaces the Free=1 and paid=unlimited quantities with a universal maximum of 10.

## Capacity and cost target

Keep routine delivery within Resend's current free tier (3,000 emails/month and 100/day), while reserving headroom for verification and password-reset mail.

Current production snapshot (2026-08-31):

- 357 users; 349 verified
- 106 users active in the last 30 days
- 120 watchlists owned by 105 users
- 36 watchlist owners active in the last 30 days
- 0 watchlists currently have alerts enabled

At the current 105 watchlist owners, one digest per owner per week is about 455 emails/month. Even all 357 accounts would be about 1,547/month, but a single weekly blast could exceed the daily free-tier cap.

## Product behavior

- Alerts are opt-in per watchlist and default off.
- Remove the paid-subscription gate from the existing alerts toggle.
- Label the control as weekly email notifications; do not expose cadence selection yet.
- Send only to verified users.
- Send at most one consolidated digest per user per delivery period.
- Include only active jobs first seen since that user's last successfully processed window.
- Use current watchlist filters at send time.
- Deduplicate jobs that match multiple watchlists and show which watchlists matched.
- Cap the email at 20 jobs, with links back to relevant watchlists for the full result set.
- Do not send an email when there are zero matches; record the skipped window so it does not get scanned repeatedly.
- Localize subject and body for all supported locales and safely escape dynamic content.
- Include visible notification-management and unsubscribe controls, including one-click unsubscribe where supported.

## Global notification pause

Add a global override in Settings that lets a user pause all email notifications without changing their individual watchlist preferences.

Effective delivery eligibility must follow this precedence:

```ts
const shouldNotify =
  watchlist.alertsEnabled && !userPreferences.notificationsPaused;
```

[...]

PR #8386, opened 2026-08-31
Add providerless weekly notification scheduler core

viktor-shcherb:
## Summary

- assign each user a deterministic UTC cadence slot across stable weekday/hour buckets, with explicit half-open sweep and match windows
- select only verified, unpaused owners with currently enabled watchlists, batch-load company scope and ledger state, and invoke the merged canonical matcher with each watchlist's enable/resume/processed floor
- claim durable delivery identities with uniqueness plus a stale-pending lease, retry definitive failures/due quota deferrals, and fail closed for ambiguous `unknown` outcomes
- aggregate and deduplicate matches while retaining watchlist labels, record unique and per-watchlist totals, and cap providerless display plans at 20 jobs
- advance zero-match windows, calculate caller-supplied daily/monthly quota decisions, bound in-process work, and return aggregate-only telemetry
- keep execution strictly `off`/`shadow`; `off` performs no reads and this branch has no provider callback, route, cron, Redis lock, environment reader, email renderer, or deployment activation

## Boundaries

This is the focused scheduler-core subset of PR C requested for #8317 / #8366. Runtime cron protection/locking and live provider delivery remain deliberately impossible and out of scope here, alongside UI/copy, email rendering/unsubscribe, paid/cap changes, and external configuration.

The human UI taste gate is not applicable because this PR changes no rendered UI, interaction, email rendering, or product copy. The human cost gate is not applicable because caps are caller-supplied pure inputs and this PR changes no provider, tier, reserve, cost envelope, or live configuration.

## Evidence

- `pnpm --dir apps/web exec vitest run src/lib/notifications/__tests__/policy.test.ts src/lib/notifications/__tests__/scheduler-policy.test.ts src/lib/notifications/__tests__/scheduler-core.test.ts` — 21 passed
- `pnpm --dir apps/web exec tsc --noEmit --pretty false` — passed
- pre-commit web ESLint and Lingui translation coverage — passed

References #8317
References #8366

Do not merge as part of this task.


apps/web/src/lib/notifications/scheduler-core.ts
```diff
@@ -0,0 +1,392 @@
...
+    quota: NotificationQuotaState;
+    concurrency: number;
+  },
+  dependencies: NotificationSchedulerDependencies,
+): Promise<{
+  plans: readonly NotificationDeliveryPlan[];
+  telemetry: NotificationRunTelemetry;
+}> {
+  const mode = input.mode ?? DEFAULT_NOTIFICATION_EXECUTION_MODE;
+  const startedAt = dependencies.now?.() ?? new Date();
```
viktor-shcherb, line 284: The documented off-mode contract is zero reads/side effects, but the injected `now` dependency is invoked before the off return. A clock hook is still an externally supplied callback and may itself read or mutate state; the current test does not spy on it. Return the empty off result before consulting `dependencies.now`, and assert that every dependency hook remains untouched.

apps/web/src/lib/notifications/scheduler-core.ts
```diff
@@ -0,0 +1,392 @@
...
+    return [{
+      user,
+      scheduledFor: user.openDelivery.scheduledFor,
+      originalWindow: {
+        windowStart: user.openDelivery.windowStart,
+        windowEnd: user.openDelivery.windowEnd,
+      },
+    }];
+  }
+  const [scheduledFor] = getNotificationScheduleSlots({
```
viktor-shcherb: `getNotificationScheduleSlots` explicitly supports recovery sweeps spanning more than one cadence period and returns every half-open slot, but this destructuring silently keeps only the first. Re-running the same broad sweep then hits the first durable identity again and never reaches later slots. Either emit one work item per returned slot with bounded recovery semantics, or reject/limit sweeps to the single-slot contract and update the policy documentation/tests.

apps/web/src/lib/services/notification-scheduler.ts
```diff
@@ -0,0 +1,395 @@
...
+} from "@/lib/watchlist-matcher-contract";
+
+const COMPLETED_STATUSES: NotificationDeliveryStatus[] = ["sent", "skipped"];
+
+function identityKey(userId: string, cadence: NotificationCadence): string {
+  return `${userId}\u0000${cadence}`;
+}
+
+async function listEligibleUsers(): Promise<readonly EligibleNotificationUser[]> {
+  const rows = await db
```
viktor-shcherb: This query materializes every eligible watchlist and every company membership before it can determine which users are in the current cadence bucket, then expands every owner into two unbounded `IN (...)` ledger queries below. There is no page size, owner cap, due-bucket predicate, or keyset loop; the in-process matcher concurrency limit does not bound this database/memory fan-out and the parameter list can eventually exceed practical/Postgres limits. Add a bounded/keyset due-owner selection and bounded batch loading, with a test asserting query/batch bounds.

viktor-shcherb (commented):
Blocking technical review: three actionable findings remain. GitHub cannot record REQUEST_CHANGES on this self-authored PR, so this COMMENT review is not approval.

The providerless boundary is otherwise clean: the diff adds no route, cron, Redis lock, provider callback/import, environment reader, UI, renderer, or cost/live activation. The repository filters for verified, unpaused owners and enabled watchlists; the core uses the canonical matcher, retains deduplicated watchlist labels, caps display output at 20 while retaining aggregate totals/truncation, records zero-match windows as skipped, uses unique delivery identities plus lease fencing, and fails closed for pre-existing `unknown` outcomes. Quota calculation is pure caller-snapshot policy and explicitly performs no reservation, so live atomic quota enforcement remains future caller scope.

Checks at `4012d621`: `git diff --check` passed; focused Vitest passed (4 files, 23 tests, including canonical matcher tests); targeted ESLint passed; `pnpm db:check` passed; full web `pnpm typecheck` passed. No scheduler, migration, provider, or production command was executed.

apps/web/src/lib/notifications/scheduler-core.ts
viktor-shcherb: Fixed in commit 50a96625d: off mode now returns the empty result before reading the clock or invoking any repository or matcher callback. The test uses a clock spy and asserts the clock, candidate-page read, hydration read, and matcher are all untouched.

apps/web/src/lib/notifications/scheduler-core.ts
viktor-shcherb: Fixed in commit 50a96625d: recovery no longer destructures the first slot. Every returned slot is processed chronologically per user, zero or completed windows advance the in-memory floor, and a non-advancing outcome stops later periods. The slot function rejects recovery beyond the explicit eight-slot bound instead of truncating. Tests cover two-period recovery order and count plus the hard recovery bound.

apps/web/src/lib/services/notification-scheduler.ts
viktor-shcherb: Fixed in commit 50a96625d: eligibility is now a keyset-paged verified and unpaused owner query (50 owners plus one lookahead) using an existence predicate for enabled alerts without materializing memberships. Cadence slots are filtered in core before hydration. Only due owner IDs (maximum 50) enter ledger queries and watchlist hydration; watchlists and company memberships use stable bounded pages (250 and 500). The result exposes an owner continuation cursor. A 60-owner test proves the 50 and 10 page and hydration bounds.

viktor-shcherb (commented):
Follow-up technical review: the off-mode callback issue is fixed, and owner queries now use a stable bounded keyset page with due-only watchlist hydration. Three continuation/bounding issues still prevent approval. GitHub cannot record REQUEST_CHANGES on this self-authored PR, so this COMMENT review is blocking and is not approval.

Positive verification: off mode returns before clock/repository/matcher callbacks and the test now asserts all remain untouched; recovery slot generation is half-open, chronological, capped at 8, and rejects rather than truncates oversized sweeps; owner IDs and ledger `IN` lists are capped at 50; watchlist and membership SQL reads use keyset/chunk pagination; canonical matcher/providerless boundaries remain unchanged. The diff adds no provider, route, cron, Redis, environment reader, UI, renderer, or cost activation.

Checks at `50a96625`: `git diff --check` passed; focused Vitest passed (4 files, 26 tests); targeted ESLint passed; `pnpm db:check` passed; full web `pnpm typecheck` passed. No scheduler, provider, migration, or production command was executed. Required CI still governs merge.

apps/web/src/lib/services/notification-scheduler.ts
```diff
@@ -0,0 +1,461 @@
...
+  return {
+    postings: ordered,
+    uniqueMatchCount: ordered.length,
+    watchlistMatchCount,
+    truncated,
+  };
+}
+
+/** Providerless entry point; no route or runtime configuration invokes it. */
+export async function runNotificationScheduler(input: {
```
viktor-shcherb, line 447: The core now returns an owner-page continuation, but the only concrete service entry point neither accepts nor forwards `cursor`. Every `runNotificationScheduler` call therefore starts after `null`; once more than 50 eligible owners exist, owners 51+ cannot be reached through the exported scheduler and starve on every run. Add cursor input/forwarding (and document or provide the bounded continuation loop), then test through this wrapper rather than only calling the core directly.

apps/web/src/lib/notifications/scheduler-core.ts
```diff
@@ -0,0 +1,486 @@
...
+          continue;
+        }
+        if (
+          outcome.kind === "duplicate" &&
+          advancesNotificationWindow(outcome.status)
+        ) {
+          lastProcessedWindowEnd = outcome.scheduledFor;
+          continue;
+        }
+        break;
```
viktor-shcherb, line 489: `not_due` falls through to this `break`. In a recovery sweep, a user can have two slots with `alertsEnabledAt` or the resume floor between them: the first slot correctly returns `not_due`, but the later slot is eligible and is never processed. Treat `not_due` as a safe continuation without advancing the processed floor, and add this boundary test; only nonterminal/failed/conflicted outcomes should block later slots.

apps/web/src/lib/services/notification-scheduler.ts
```diff
@@ -0,0 +1,461 @@
...
+      .innerJoin(userPreferences, eq(userPreferences.userId, watchlist.userId))
+      .where(and(
+        inArray(watchlist.userId, [...userIds]),
+        eq(watchlist.alertsEnabled, true),
+        isNotNull(watchlist.alertsEnabledAt),
+        watchlistCursor ? gt(watchlist.id, watchlistCursor) : undefined,
+      ))
+      .orderBy(asc(watchlist.id))
+      .limit(NOTIFICATION_WATCHLIST_QUERY_PAGE_SIZE);
+    watchlistRows.push(...page);
```
viktor-shcherb: These SQL page limits bound each query, but not total hydration: every page is appended to `watchlistRows`, then every membership page is retained in `companyIds` before matching starts. Grandfathered owners above the current watchlist ceiling remain supported, and company memberships have no hard cardinality cap, so one 50-owner page still has unbounded memory/work fan-out. Process smaller bounded owner/watchlist hydration units through the core, stream/chunk matching, or enforce and test an actual total hydration bound; a page-size constant alone is not that bound.

apps/web/src/lib/services/notification-scheduler.ts
viktor-shcherb: Fixed in commit 782891d98: the exported service entry point now accepts the owner cursor and forwards it unchanged to the core. A service-level test follows the first result continuation into the second call and asserts user-049 is forwarded, while the 60-owner core test proves owners 50 through 59 are reached without starvation.

apps/web/src/lib/notifications/scheduler-core.ts
viktor-shcherb: Fixed in commit 782891d98: not_due is now a safe bounded-scan continuation that does not move the processed floor. Only non-advancing delivery outcomes stop later periods. A recovery test places the alert-enable floor between two weekly slots and proves the first is skipped while the second is claimed and processed.

apps/web/src/lib/services/notification-scheduler.ts
viktor-shcherb: Fixed in commit 782891d98: hydration no longer accumulates a 50-owner page. Each due period cursor-streams one watchlist and at most 100 company memberships per segment, with a 500-segment hard budget and a 2,500-posting aggregate memory cap. It scans segments to establish the durable floor, claims once, then matches streamed segments. The stress test covers 12 grandfathered watchlists with 225 memberships each, reaches all 12, performs 36 match segments, and proves no segment exceeds 100 memberships.

viktor-shcherb (commented):
Final technical re-review: the prior owner-cursor starvation, `not_due` recovery starvation, unbounded all-owner hydration, and off/providerless issues are fixed. The service forwards the owner cursor, the 60-owner test reaches the second page, recovery continues without advancing the floor past a pre-enable slot, hydration streams one watchlist / at most 100 memberships at a time, aggregation retains at most 2,500 postings, and the diff still adds no provider, route, cron, Redis, environment reader, UI, renderer, or live/cost activation. One hard-limit boundary defect remains, so this COMMENT review is blocking rather than approval (GitHub cannot record REQUEST_CHANGES on this self-authored PR).

Checks at `782891d`: `git diff --check` passed; focused Vitest passed (5 files, 29 tests); targeted ESLint passed; `pnpm db:check` passed; full web `pnpm typecheck` passed. Required CI was not bypassed, and the PR remains open.

apps/web/src/lib/notifications/scheduler-core.ts
```diff
@@ -0,0 +1,575 @@
...
+        cursor,
+      });
+      if (page.watchlist) await visitor(page.watchlist);
+      if (!page.nextCursor) return;
+      if (JSON.stringify(page.nextCursor) === JSON.stringify(cursor)) {
+        throw new Error("Notification hydration cursor did not advance");
+      }
+      cursor = page.nextCursor;
+    }
+    throw new Error("Notification hydration segment budget exhausted");
```
viktor-shcherb: The 500-segment boundary is off by one. The concrete repository returns a non-null advance cursor after every real segment—including the final watchlist/membership segment at service lines 243-253—and only reports EOF on the following empty read. Consequently, 500 real segments consume all 500 loop iterations and this throws before that terminal probe; such an owner is reported `hydration_failed` forever even though it is exactly at the declared cap. The new stress test exercises only 36 segments and returns zero postings, so it covers neither this 500 boundary nor the 2,500 aggregate boundary. Permit an EOF-only lookahead without visiting a 501st data segment (or make the repository mark the final segment terminal), and add boundary tests for 500 vs. 501 data segments plus 2,500 vs. 2,501 unique postings; the membership test should also assert all IDs appear exactly once.

viktor-shcherb: Fixed in commit b03753554. The cursor loop now counts real data segments separately from bounded cursor probes: exactly 500 visited segments may perform one EOF-only lookahead, while a real segment 501 is rejected before its visitor, claim, or matcher runs. New non-empty boundary tests prove 500 segments succeed with 500 distinct memberships and postings consumed exactly once; 501 fails before matching; 2,500 unique postings are retained without truncation; and posting 2,501 sets truncation while retaining the bounded 2,500-item aggregate and the full 2,501 occurrence total.

viktor-shcherb (commented):
Final technical approval at `b0375355` (recorded as COMMENT because GitHub cannot approve a self-authored PR). No blocking findings remain.

The hydration walk now separates 500 visited data segments from a bounded 1,001-probe budget, so exactly 500 real segments complete after the terminal lookahead, while a real segment 501 throws before the first-pass visitor can reach claim or matching. The tests exercise both boundaries and assert the 501 case has no claim/matcher calls. The 2,500/2,501 aggregate cases prove exact-cap retention and bounded truncation, and the 12 x 225 membership stream asserts all 2,700 non-empty membership IDs plus generated posting IDs are observed exactly once. Owner cursor forwarding beyond 50, pre-enable `not_due` recovery, off-mode zero callbacks, bounded recovery, durable/fail-closed delivery handling, and canonical matcher behavior remain covered. The PR remains providerless and inactive: no provider callback/import, route, cron, Redis lock, environment reader, UI/copy, renderer, or cost/live configuration is added.

Evidence: full PR `git diff --check` passed; focused Vitest passed (5 files, 33 tests); targeted ESLint passed; `pnpm db:check` passed; full web `pnpm typecheck` passed. Merge remains conditional on unchanged head, required CI green, and ordinary repository protections—no bypass.

Issue #8401, opened 2026-09-01
[crawler-roi] Repair process-tree cadence, coherent snapshots, and browser retry evidence


viktor-shcherb:
Parent capture: #8228
ROI parent: #7936
Provenance only: closed #8159 / merged #8161
Rejected predecessor capture: #8171

## Outcome

Repair the production telemetry that made the #8228 old-Python baseline a hard NO-GO. The repaired release must sample each crawler role at a deadline-anchored fixed rate, expose root and process-tree CPU/RSS from one coherent observation, account for full initial cgroup CPU, and make every bounded browser retry family observable at zero or above. Then deploy normally and pass a passive 15–30 minute six-target burn-in before #8228 may request a new independent GO.

This is a successor repair. #8159 is closed and #8161 is merged provenance; neither is an active owner. #8171 records a rejected historical window and does not own this change.

## Production failure being repaired

The clockless preflight and independent audit are authoritative:

- owner NO-GO: https://github.com/colophon-group/jobseek/issues/8228#issuecomment-5487413200
- independent audit: https://github.com/colophon-group/jobseek/issues/8228#issuecomment-5487438037

On crawler `0.13.636`, `browser-1` accrued about 143 declared sampling gaps in five minutes and achieved only about 70% of the expected 0.5-second observations. HTTP-worker tree RSS intermittently fell below root RSS. Exporter/drain tree CPU remained below root CPU because the first cgroup observation established zero after root startup accounting had begun. Browser navigation-network and target-closed retry series were absent; absence is unknown, never observed zero.

No current repair PR exists. The #8228 deployment hold remains active and its 86,400-second clock remains unset.

### Absolute-deadline fixed rate

- Use a monotonic clock and one immutable interval `I` with deadlines `D(n) = D(0) + n*I`. Do not sleep for `I` after work completes.
- Publish at most one completed observation per scheduled deadline. Never run an unbounded catch-up burst.
- When a collection finishes after one or more subsequent deadlines, increment the sampling-gap counter by the exact number of skipped deadlines and advance to the first future deadline. Traversal time must not cause phase drift or be silently added to the declared interval.
- A collection error increments the failure outcome exactly once. Skipped deadlines increment gaps exactly; failure and gap accounting must not substitute for or double-count one another.
- The configured interval remains positive and no greater than one second. Deadline arithmetic and counters are monotonic across ordinary clock adjustments because wall clock is not the cadence source.

### Same-observation root/tree accounting

[...]

Issue #8402, opened 2026-09-01
[crawler-roi] Capture bounded Chromium transport and routing denominator before Python baseline


viktor-shcherb:
Parent capture: #8228
ROI parent: #7936
Shared-HTTP provenance: merged #8232
Backend-aware ROI contract: #8239

## Outcome

Close the browser-transport evidence gaps before the exact old-Python baseline starts. Instrument Chromium so one fixed-window capture can conserve bounded main-document, redirect, subresource, and warmup attempts/outcomes, transferred response bytes, actual direct/proxy route, and registry-backed capability/provider attribution without creating high-cardinality labels.

The deliverable is capture-ready telemetry and fail-closed adapter/schema/model evidence. It does not start #8228, calculate ROI, generate projected traffic, or claim Lightpanda equivalence.

## Evidence gap being closed

The production preflight and independent audit are authoritative:

- owner NO-GO: https://github.com/colophon-group/jobseek/issues/8228#issuecomment-5487413200
- independent audit: https://github.com/colophon-group/jobseek/issues/8228#issuecomment-5487438037

Merged #8232 measures SSRF-approved traffic through the crawler shared HTTP client. Its `execution_class=browser` is still shared-HTTP traffic; it does not measure Chromium page navigation, redirect hops, subresources, or warmup traffic. Therefore current browser origin attempts, bytes, response-size distribution, direct/proxy route, and provider-weighted load remain unknown. Capability counters alone do not freeze the routed workload denominator.

No current repair/implementation PR exists. The #8228 deployment hold remains active and its 86,400-second clock remains unset.

### Attempt and outcome boundary

- Count one Chromium network attempt when a real browser transport request is admitted to its selected route. A request rejected before transport admission is not an attempt and must use a separate bounded pre-transport outcome.
- Classify every attempt into exactly one closed request class: `main`, `redirect`, `subresource`, or `warmup`.
  - `main`: the initial top-level document request for the accepted browser task.
  - `redirect`: each subsequent HTTP redirect hop, counted as its own attempt.
  - `subresource`: a non-top-level request initiated by the page after navigation admission.
  - `warmup`: an explicitly configured browser/service readiness request that is not task work; it must remain separately attributable and may not inflate task success.
- Every admitted attempt reaches exactly one terminal bounded outcome. The frozen outcome enum must distinguish at least complete response, partial response, no-response transport failure, policy rejection after admission, cancellation, and target closure. One attempt cannot be conserved into multiple outcomes.
[...]

Issue #8228, opened 2026-08-30
[crawler-roi] Retry exact 24h Python process-tree capture under deployment hold


viktor-shcherb:
Parent: #7936
Failed predecessor: #8171

## Outcome

Produce one valid, exact 86,400-second Python crawler process-tree capture without any release transition. The predecessor `.593` window is rejected and is not reusable: automatic `.594` deployment changed the six-target fleet roughly 1h02m after its start.

This issue owns only capture governance, preflight, sanitized read-only snapshots, and the final aggregate. It does not deploy, restart, mutate infrastructure, compute sustainable totals, or calculate Go ROI.

## Current release queue — do not start yet

Fresh main at reservation is `7de7a4975367c0e06eab1b49f96368e029c5d4af`, crawler `0.13.598`. Open PRs currently reserve/touch crawler VERSION through `.607`, including duplicate/stale reservations: #8174 `.594`, #8184 `.599`, #8192 `.600`, #8198 `.601`, #8200 `.602`, #8201 `.603`, #8211 `.604`, #8217 `.605`, #8218 `.606`, #8223 `.607`, and #8225 `.599`.

The next window must not start while any `.599`–`.607` owner is expected to merge/deploy or while any crawler deployment is queued/running.

## Existing enforceable hold

Reuse the repository deployment governance; do not invent a bypass:

1. When the release queue is drained, apply the existing `deployment-hold:crawler` label to this open issue.
2. Wait for `.github/workflows/crawler-deploy-gate.yml` to reconcile every open PR and verify ready runtime PRs are blocked at their exact heads.
3. Confirm there is no queued/running `Deploy Crawler (Hetzner)` run and that main VERSION equals the unique six-target live `crawler_build_info` version.
4. Keep the hold issue open and labeled through end acceptance. No direct push, admin override, or manual infrastructure/deploy action is authorized.
5. Remove the label only after the final sanitized end snapshot is accepted or the window is explicitly aborted.

The label is intentionally **not applied at issue creation** because the active version queue must settle normally first.

## Exact preflight and automation contract

Before choosing a boundary:

- six unique targets: worker-1/2/3, browser-1, exporter, drain;
- one identical exact release on all six and matching current main VERSION;
- all required process-tree families present, interval <=1 second, heartbeat/boundary freshness <=60 seconds;
- both sample outcomes plus CPU/RSS/descendants/gaps/sampler-start counters present;
- tree RSS >= root RSS and no preflight counter anomaly;
[...]

PR #8407, opened 2026-09-01
Isolate process-tree sampler cadence

### LINKED ISSUE #8405: [crawler-roi] Isolate process-tree sampler cadence from browser interpreter stalls
## Relationship and release sequence

This is the single bounded repair sub-issue for #8401. It blocks the clean production burn-in and independent acceptance required to close #8401.

#8228 remains blocked by #8401 and this repair. Its 24-hour measurement clock stays unset, and the deployment hold remains active. Do not start that clock until this issue is implemented, deployed, burned in, and independently accepted through #8401.

Authoritative failed burn-in: https://github.com/colophon-group/jobseek/issues/8401#issuecomment-5489004831

## Proven diagnosis

The clean 1,200.004984 second production interval on crawler 0.13.637 showed browser-1 at 1,757 successful samples out of 2,399 expected opportunities, plus 643 counted gaps. All five other long-lived processes were near or at 100 percent with zero gaps, and browser-1 had zero sample failures, starts, or resets. The container did not restart or OOM, the release identity was exact, root and tree pairs remained atomic, tree was never below root, and the retry evidence matrix was complete.

Additional fixed-window evidence establishes the failure category:

- Browser-1 counter conservation was exact: 3,104 successes plus 496 gaps equaled 3,600 deadlines. This is real deadline loss, not clock drift or counter arithmetic loss.
- Gap-heavy 15 second bins had much higher crawler root Python CPU but lower total process-tree CPU, fewer descendants, lower host load, and lower task and scrape throughput. Correlation with gap count was +0.856 for root CPU and -0.904 for tree CPU. This contradicts Chromium count, process-tree scale, and aggregate host load as the primary cause.
- A bounded production schedstat observation put sampler execution near 11 ms of CPU per sample, far below the 500 ms interval. Process collection duration is therefore not the primary overrun mechanism.
- Existing sample reads are defensive; disappearing proc entries would surface failures, but production failures were zero. The Prometheus collector copies one immutable snapshot under a short lock and the exporter is a separate container.
- Memory-reclaim events were negatively correlated with gaps, and isolated stress of common normalization, language detection, and technology matching paths did not reproduce deadline loss.
- A deterministic isolation probe reproduced the boundary: a native sleep that released the Python GIL produced zero gaps, while the same sleep through PyDLL, which held the parent interpreter GIL, produced counted gaps. A dedicated Python thread is therefore not isolated from parent interpreter stalls.
- CPU quota and pressure are present and may amplify wake latency, but gap-heavy bins had lower aggregate tree and host load, so they are secondary rather than the leading explanation.

High-confidence root-cause category: the in-process sampler thread suffers wake-to-run starvation while browser root-main-thread work monopolizes the Python interpreter. Parent GIL starvation is the dominant demonstrated boundary, with cgroup scheduling pressure as a possible amplifier.

The exact synchronous production primitive is not yet observable. Current telemetry has no scheduled-start, actual-start, collection-end, or publish-end timestamps, so this issue must not claim a specific browser function or reinterpret all gaps as collection overruns.

## Required implementation

1. Move cadence and process-tree collection out of the browser parent interpreter into one minimal independently scheduled sampler process. Keep it in the same container and cgroup, observe the exact crawler root PID and its descendants, and preserve current cgroup CPU authority. A second Python thread is not sufficient.
2. Publish only complete snapshots over an atomic IPC boundary. Each snapshot must include a monotonically increasing sequence, monotonic sample timestamp, root and tree CPU, root and tree RSS, descendant count, and the existing CPU source identity. The parent collector must never combine fields from different sequences.
3. Treat sampler child death, stale sequence, malformed snapshot, and IPC failure as fail-closed sample failures. Never infer zeros. Restarts and baseline resets must remain explicit and observable.
4. Preserve absolute monotonic deadline scheduling and no-catch-up semantics. For every skipped deadline, classify it exactly once as scheduler-late if it elapsed before collection began, or collection-overrun if it elapsed during collection or handoff. The sum of reason counters must equal the existing total gap counter with no double counting.
5. Add bounded, low-cardinality timing evidence for wake lateness, collection duration, and handoff or publish duration. Do not add company, board, URL, posting, task, or process-ID labels.
6. Document that sampler-process overhead is bounded and included in the observed tree totals. Preserve the root and tree invariants and the initial nonzero cgroup CPU baseline contract.
7. Keep capture and model fail-closed on any gap, failure, start, or reset. Do not relax acceptance, hide a stall by changing the interval, or substitute workload, queue, browser concurrency, CPU quota, or host changes for the isolation repair.

Expected ownership is limited to the runtime-cost sampler and its metric handoff, the measurement schema only if timing evidence is captured there, focused tests, the runtime-cost documentation, and VERSION. Before implementation, freeze an exact writable-file allowlist in the repair PR. Do not reopen or modify #8402.

## Deterministic test acceptance

The repair is not complete without tests that prove all of the following:

- A controlled parent GIL hold longer than 500 ms does not make the isolated sampler miss its 500 ms cadence. The test must demonstrate that the old in-process-thread boundary would miss.
- Scheduler lateness and collection or handoff overrun are injected independently and classified exactly, with total-gap conservation and no catch-up burst.
- Concurrent reads cannot observe a torn or mixed-sequence snapshot.
- Child death, restart, stale sequence, malformed IPC, and unavailable snapshot all fail closed and surface explicit failure, start, or reset evidence as appropriate.
- Initial nonzero cgroup CPU, root-to-tree accounting, tree greater than or equal to root, and atomic CPU and RSS pairing remain intact.
- High process count, descendant churn, and disappearing proc entries stay within the deadline or produce explicit failures rather than silent gaps.
- Simulated 86,400 second schedules at both 0.5 second and 1.0 second intervals conserve every deadline exactly, including monotonic clock jumps supported by the current scheduler contract.
- Container packaging, process cleanup, signal handling, and image/runtime parity include the new sampler process.

Run the focused runtime-process-tree, metrics-server, and runtime-cost suites plus the full crawler test suite and repository CI.

## Deploy and production burn-in acceptance

- Ship through an ordinary reviewed PR with a new monotonic crawler version, green full CI, runtime-contract checks, CodeQL, and image/runtime parity.
- Deploy only through the guarded normal path while the #8228 measurement hold is active; restore and verify the hold after deployment. Do not start the #8228 clock.
- Use one fresh production-only interval of at least 30 continuous minutes after all six long-lived processes report the exact new release and uptime wholly covering the interval.
- During that interval, all six processes must have zero process-tree sample gaps, failures, unexpected starts, baseline resets, container restarts, and OOMs. The canonical counter query must conserve every deadline, and successful coverage must be at least 95 percent for every process.
- Browser-1 timing evidence must show wake-lateness p99 below 100 ms, collection-duration p99 below 100 ms, handoff-duration p99 below 50 ms, and every observed value below 250 ms. Any deadline at or above 500 ms is an automatic failure regardless of percentile.
- Preserve atomic root and tree CPU and RSS pairs, tree greater than or equal to root, nonzero cgroup baselines, and the complete browser retry evidence matrix.
- Do not change the sampling interval, browser concurrency, queues, workload mix, CPU quota, or host size during the acceptance interval.
- Post raw query windows, release and uptime proof, restart and OOM proof, reason-counter conservation, timing distributions, process-tree invariants, and retry evidence to #8401 for independent acceptance.

Only after #8401 accepts this repair and its clean burn-in may #8401 close and #8228 choose a fresh 24-hour clock start.

viktor-shcherb:
## Summary

- isolate process-tree sampling in a spawned same-container/cgroup process so parent-interpreter GIL stalls cannot hide browser-worker CPU
- publish immutable cumulative evidence over bounded atomic Unix datagrams and preserve monotonic offsets across supervised child restarts
- classify skipped deadlines causally after the complete IPC handoff as `scheduler_late` or `collection_overrun`, with exact conservation and bounded timing histograms
- reject malformed, future, reordered, and regressing frames without replacing the latest sample or refreshing parent-local stale recovery

This is the independently approved exact head `f294593a2468970a020934885c009f1436bc4f71` against main base `168b72b88b12b9e2d25b8414cec92901e007be85`. Approval: https://github.com/colophon-group/jobseek/issues/8405#issuecomment-5489677927

## Local verification

- Ruff check and format check: passed
- Pyright: 0 errors, 0 warnings
- process-tree suite: 37 passed
- runtime/metrics/contract focused suite: 644 passed
- full crawler suite: 11,564 passed, 82 expected environment skips, 8 existing warnings
- runtime-v1 deterministic generation/protocol/compatibility/privacy/semantics: passed
- runtime conformance: 497 passed
- isolated `jobseek-crawler==0.13.638` wheel build/install, packaged sampler/collector import, and all six installed-role build-info smokes: passed

## Mandatory hosted gates before merge

- [ ] slim image build and installed-image smoke
- [ ] full/browser image build and installed-image smoke
- [ ] Linux sampler smoke proving the sampler child and exact crawler root share the expected container/cgroup, and that sampling observes the intended root PID/tree
- [ ] SIGTERM lifecycle smoke proving graceful shutdown leaves no sampler orphan, plus bounded forced container-stop observation under the reviewed init/grace configuration
- [ ] full hosted test suite and runtime-contract suite on this exact commit

The deployment-hold check is expected to remain red. Do not remove, waive, or reinterpret that hold as approval to merge or deploy. Hosted failures require classification and independent review.

This PR does not touch #8402 transport/routing/contracts and must not start #8228 burn-in. It closes implementation issue #8405 only; it explicitly does **not** close parent #8401, which remains open until the separately authorized production burn-in and evidence review are complete.

Closes #8405


viktor-shcherb:
## Immutable promotion evidence

Promotion guard passed before push:

- fetched `origin/main`; it remained exact reviewed base `168b72b88b12b9e2d25b8414cec92901e007be85`
- remote PR head is exact approved commit `f294593a2468970a020934885c009f1436bc4f71`
- direct reviewed parent is `69cf3e8754beda6c4e8b4fd1117b9a0523ba02b0`
- clean worktree and `git diff --check` passed
- aggregate delta is exactly the approved six-file allowlist; `VERSION` is `0.13.638`; no #8402 surface
- independent corrected-head approval: https://github.com/colophon-group/jobseek/issues/8405#issuecomment-5489677927

Local exact-head evidence: Ruff/format/Pyright passed; process-tree 37/37; focused runtime/metrics/contracts 644 passed; full crawler 11,564 passed with 82 expected environment skips and 8 existing warnings; runtime conformance 497 passed; isolated 0.13.638 wheel install/import and six-role build-info smoke passed.

Hosted Linux/image/lifecycle checks listed in the PR description remain mandatory. The deployment hold is expected to remain red and is not to be waived. This evidence does not authorize merge, deploy, #8401 closure, hold removal, or #8228 burn-in.

viktor-shcherb:
## Terminal hosted-check classification\n\nReviewed commit: f294593a2468970a020934885c009f1436bc4f71  \nBase commit: 168b72b88b12b9e2d25b8414cec92901e007be85  \nIndependent approval: https://github.com/colophon-group/jobseek/issues/8405#issuecomment-5489677927\n\nAll hosted jobs are terminal. There are no remaining jobs.\n\n### Passed\n\n- Approval gates 1 and 2: the Linux amd64 image job built the slim crawler image and the full/browser image, verified installed build information for all six long-lived roles, tested the installed sync data-root contract, and smoke-tested the full image read-only entry point: https://github.com/colophon-group/jobseek/actions/runs/33476584829/job/99757145782\n- Approval gate 5: full crawler tests, PostgreSQL and Typesense E2E, Python lint/type, version check, coverage, web checks/build, dependency/security checks, and Required CI are green. Full crawler job: https://github.com/colophon-group/jobseek/actions/runs/33476584829/job/99757146023\n- Runtime bindings/conformance, protocol corpora, Go external-consumer conformance, and isolated crawler-wheel checks are green: https://github.com/colophon-group/jobseek/actions/runs/33476584827/job/99757009266\n- Installed image parity is green: https://github.com/colophon-group/jobseek/actions/runs/33476584827/job/99757291021\n- Required CI is green: https://github.com/colophon-group/jobseek/actions/runs/33476584829/job/99758767007\n\n### Expected policy result\n\n- Crawler Deploy Gate is red because crawler deployment is held: https://github.com/colophon-group/jobseek/actions/runs/33476897809\n- This is the expected policy hold, not an implementation or CI failure. No action was taken to release it.\n- Probe new/changed boards was skipped because this PR has no board changes; it is irrelevant to #8405.\n\n### Mandatory evidence still missing\n\nThe hosted workflows did **not** execute the following approval-specific live-container gates:\n\n1. A real Linux installed-image sampler run proving that the sampler observes the exact crawler root PID, remains in the same container/cgroup, and reports the crawler-root process tree.\n2. A real container lifecycle run proving SIGTERM leaves no sampler orphan, plus forced container stop behavior under init/grace.\n\nThe passing unit tests use deterministic or synthetic proc/cgroup fixtures, and the installed-image smoke jobs do not start and inspect the live sampler child under container init. They therefore do not substitute for gates 3 and 4.\n\n### Classification\n\nThere is no unexpected code or CI failure. The PR is correctly OPEN and BLOCKED. It **must not merge or deploy** until gates 3 and 4 are added/run against this exact reviewed commit and their evidence is independently reviewed. The deployment hold remains in place, and #8401 remains open for post-merge burn-in.

viktor-shcherb:
CHANGES REQUIRED — hosted lifecycle gates 3/4 are still missing at exact PR head `f294593a2468970a020934885c009f1436bc4f71`.

The existing hosted `crawler-image` check is useful but insufficient: it builds the slim/full images and exercises installed import/help only. It never starts the real crawler roles, so it cannot prove that `ProcessTreeSampler` is attached to the intended crawler root inside the production image topology, nor that either graceful or forced container shutdown leaves no sampler orphan.

## Frozen smallest durable scope

Exact file allowlist:

1. `.github/scripts/crawler-sampler-container-smoke.py` — new reusable, standard-library host-side Docker lifecycle probe.
2. `.github/workflows/ci.yml` — extend the existing required `crawler-image` job; reuse its already-built exact slim/full images.
3. `scripts/ci-workflow.test.mjs` — static contract asserting that the required image job invokes this probe for both images with the pinned services/provenance/evidence requirements below.

No crawler Python, Dockerfile, entrypoint, Compose, runtime-contract workflow, VERSION, or docs changes are in scope. If the smoke exposes a real lifecycle defect, stop and re-scope that defect explicitly; do not weaken the gate.

One image is not a sufficient proxy for the other. Run both exact artifacts because the full image adds `with-xvfb`, Xvfb, Playwright/Chrome, and a distinct entrypoint:

- `jobseek-crawler-slim-ci:pr`, using its default real `crawler run`
- `jobseek-crawler-ci:pr`, using its default real `with-xvfb -> crawler run-browser`

## Hosted runner topology

Extend the existing Ubuntu `crawler-image` job with pinned healthy services:

- PostgreSQL 17: `postgres:17-alpine@sha256:742f40ea20b9ff2ff31db5458d127452988a2164df9e17441e191f3b72252193`
- Redis 8: `redis:8-alpine@sha256:978f0e01593e65eed801f2402944efcd936d43b5027e4908a7897baf88ed6241`

Expose them on runner localhost and run the crawler containers sequentially with:

```
--platform linux/amd64
--init
--cgroupns=private
--network host
--read-only
--tmpfs /tmp:rw,noexec,nosuid,nodev,size=64m
--stop-timeout 30
```

Use real runtime configuration with localhost Postgres/Redis, a unique `CRAWLER_DB_ROLE`, pool min/max 1/2, discovery/monitor concurrency 1, `SHUTDOWN_GRACE_SECONDS=2` or `3`, `PROXY_PROVIDER=none`, and one non-colliding metrics port. Run only one crawler test container at a time.

No `--privileged`, `--pid=host`, host `/proc` or cgroup mounts, added capabilities, or sudo inside the test container are required or permitted. GitHub-hosted Linux already provides Docker daemon access. Private PID/cgroup namespaces plus Docker's normal cgroup-v2 and `/proc` mounts are sufficient; the smoke must fail closed if the hosted runner cannot provide that contract. `--network host` is only for the pinned job services and mirrors the deployed topology. GitHub documents runner-job service access through mapped localhost ports: [service containers](https://docs.github.com/en/actions/tutorials/use-containerized-services/use-docker-service-containers).

## Required live assertions, for each image

Start the unmodified default command and allow at most 30 seconds for readiness. Probe from inside the container and require:

- PID 1 is `docker-init`.
- Exactly one real crawler root command is present for the expected role.
- Exactly one multiprocessing sampler child exists; its `PPid` is the exact crawler PID.
- Sampler, crawler, and PID 1 have the same exact `/proc/<pid>/cgroup` membership in a cgroup-v2 namespace.
- A live metrics scrape has at least one successful sample, zero failures, one sampler start, no initial gap reason, interval 0.5s, a positive coherent observation sequence, and non-zero root/tree CPU and RSS fields.
- Tree values are not below root values, and descendant count is at least one.
- The sampled root values are compatible with the current exact crawler `/proc` values; this, the direct-parent check, and cgroup equality establish that the intended crawler root—not PID 1 or an unrelated process—is sampled.

Record all container processes before shutdown using host PID, PPID, argv, and `/proc/<pid>/stat` start time. Start time is mandatory so PID reuse cannot create a false pass/failure.

## Gate 3: bounded graceful SIGTERM

For each image:

1. After the live assertions pass, run `docker stop --time=30`.
2. Require completion within 20 seconds (well below the Docker bound), exit code 0, and `OOMKilled=false`.
3. Require crawler logs to contain `pipeline.stopped`, `cli.shutting_down`, and `cli.stopped`.
4. Within 5 seconds, require every recorded PID/start-time identity to be gone, including sampler/resource-tracker/Xvfb/browser descendants. No sampler orphan is allowed.

## Gate 4: forced escalation/init cleanup

Start a fresh instance of each image and repeat the live assertions. Then:

1. Send `SIGSTOP` to the exact crawler PID so application cleanup cannot complete.
2. Run `docker stop --time=1`; this exercises Docker init plus the real stop escalation, rather than bypassing it with an ad hoc entrypoint.
3. Require exit code 137, `OOMKilled=false`, completion under 5 seconds, and absence of `cli.stopped`.
4. Within 5 seconds, require every recorded PID/start-time identity to be gone. No sampler, Xvfb, browser, resource tracker, or other container descendant may survive.

Docker's contract is SIGTERM followed by SIGKILL after the timeout: [docker container stop](https://docs.docker.com/reference/cli/docker/container/stop/). `--init` and stop-timeout behavior are documented under [docker container run](https://docs.docker.com/reference/cli/docker/container/run).

This is four sequential real-container runs total: graceful + forced for slim, graceful + forced for full.

## Exact artifact provenance and durable evidence

The job must prove what it actually tested:

- Checkout enough history to verify the tested build tree. Disable persisted checkout credentials.
- Label both built images with `org.opencontainers.image.revision=${{ github.sha }}`, `com.colophon-group.jobseek.pr-head=${{ github.event.pull_request.head.sha }}`, and `com.colophon-group.jobseek.base=${{ github.event.pull_request.base.sha }}`.
- For a PR merge ref, verify parent 2 is the exact PR head and parent 1 is the exact base; alternatively test a direct exact-head checkout. Do not merely print the values.
- From `docker image inspect`, record and require the OCI/source labels, exact local image content ID (`sha256:...`), platform, creation time, tested build SHA, PR head, and base SHA for both images.

The script must initialize and write a JSON evidence artifact even on failure, append a concise summary to `$GITHUB_STEP_SUMMARY`, and the workflow must upload it with the repository's pinned upload-artifact action under `if: always()`. The evidence must contain per-run timings, process identities, live metric facts, inspect results, exit/OOM state, log markers, and orphan-poll result.

The lifecycle smoke should gate actual `pull_request` runs. A manual-dispatch `pr` value must not be presented as equivalent unless that path is also changed to fetch and verify the requested PR head.

## Acceptance

This comment freezes the smallest acceptable implementation scope and evidence. Gates 3 and 4 remain red/missing until a hosted `crawler-image` run at the exact candidate head passes all four runs and publishes the digest-bound JSON artifact. The expected Crawler Deploy Gate hold is unrelated and may remain red.

viktor-shcherb:
CHANGES REQUIRED — independent review of exact candidate commit `03df72ea39b1c2ea81f78d0b85b40bd6564c128c`, whose exact parent is the approved sampler commit `f294593a2468970a020934885c009f1436bc4f71`.

The change is confined to the frozen three-path allowlist and the core lifecycle assertions are directionally correct. Two durability/security gaps still prevent approval.

## 1. The tested build context is mutable after provenance verification

`.github/workflows/ci.yml:238-274` verifies the merge commit and parents, then executes the PR-controlled `.github/scripts/crawler-sampler-container-smoke.py --self-test` before the slim image is built. The full image is built still later. A changed self-test can rewrite `apps/crawler/` after the Git proof, allowing both images to carry the expected OCI head/base/revision labels while their bytes no longer come from that Git tree. This is a real false-provenance/CI-bypass path: a PR could repair a broken candidate Docker context only in the runner workspace, pass the lifecycle gate, and leave the broken committed context to be merged/deployed.

Required correction: do not execute any PR-controlled host code between the exact-tree proof and the two builds. Put both builds directly after the pinned Buildx setup, move the probe self-test after both builds, and fail closed on a dirty tracked or untracked checkout immediately before each build (while rechecking HEAD and the two merge parents). Keep recording each exact loaded image content ID and OCI labels as implemented.

## 2. Failure/cancellation cleanup is not fail-safe

`.github/scripts/crawler-sampler-container-smoke.py:957-1107` only cleans up in `except Exception`, and only when `created` was set after `docker run` returned. This misses several required paths:

- If `docker run --detach` creates the named container but the client times out or is interrupted before line 971, `created` remains false and no removal is attempted.
- `KeyboardInterrupt` is a `BaseException`, and CI cancellation normally arrives as a process signal; there is no SIGINT/SIGTERM handler or outer active-container cleanup registry/finally.
- The return value from forced removal is ignored, so partial evidence can claim failure handling occurred without proving the container was removed.
- A readiness failure before line 988 has no host PID/start-time capture, so the cleanup path cannot prove that sampler/Xvfb descendants disappeared.

Required correction: make cleanup name-driven and unconditional in a `finally` path, independent of the `created` flag; install bounded SIGINT/SIGTERM handling so cancellation reaches the same cleanup/evidence path; best-effort capture host identities as soon as a container exists; require forced removal success and poll every captured PID/start-time identity to disappearance; persist cleanup success/failure in the partial JSON before exiting. The normal successful four-run assertions should remain unchanged.

## Verified at this exact commit

- Exact diff: only `.github/scripts/crawler-sampler-container-smoke.py`, `.github/workflows/ci.yml`, and `scripts/ci-workflow.test.mjs`.
- PostgreSQL and Redis images use the frozen digests already pinned elsewhere in this repository.
- Docker calls use argv arrays rather than a shell; workflow inputs used by the probe are constants or SHA/repository contexts, and checkout credentials are not persisted.
- Both exact loaded tags are inspected and exercised separately; no `--privileged`, `--pid=host`, host `/proc`/cgroup mounts, or added capabilities are present.
- Private cgroup-v2, PID/start-time, root-metric coherence, graceful exit/OOM/log, and forced exit/OOM/no-completion assertions match the frozen contract.

Local verification passed:

- probe self-tests: 5/5
- `node --test scripts/ci-workflow.test.mjs`: 71/71
- `git diff --check` and Python AST parse
- actionlint v1.7.12: clean
- zizmor 1.26.1: no reportable findings

Hosted execution is still mandatory after correction. The PR currently remains at `f294593a...`; candidate `03df72ea...` has not produced a hosted slim/full lifecycle artifact and cannot satisfy gates 3/4 yet.

viktor-shcherb:
APPROVE — independent re-review of exact corrective commit `b28f521100704ec757f8de802b8adac62cc9001a`, with exact parent `03df72ea39b1c2ea81f78d0b85b40bd6564c128c`.

The two blockers in the prior review are closed at this exact commit.

## Provenance blocker: closed

- The merge build SHA, exact base/head parents, and two-parent shape are verified from the clean checkout.
- After pinned Buildx setup, the slim and full builds each have their own immediately preceding guard. Each guard fails on any tracked, staged, or untracked workspace change and re-verifies HEAD, parent 1, parent 2, and absence of an extra parent.
- The next named step after each guard is the corresponding pinned build action. No PR Python/Node or repository command runs between either guard and build.
- Both exact `linux/amd64` images are built and loaded before the probe self-test. Both retain the frozen OCI revision/head/base/source labels, and the lifecycle probe independently requires those labels and the exact loaded content IDs before use.

This removes the prior path where the PR-controlled self-test could mutate `apps/crawler/` before a mislabeled build.

## Failure/cancellation cleanup blocker: closed

- Every generated container name is registered before `docker run` begins, so create timeout/interruption cannot bypass name-driven cleanup.
- A best-effort host PID/start-time capture runs immediately after creation and again in the unconditional outer cleanup; stable identities are accumulated rather than replaced.
- Lifecycle cases and the probe boundary catch `BaseException`. SIGINT/SIGTERM are converted into a cleanup-aware interruption, and repeated signals do not re-enter or abort cleanup.
- The outer `finally` force-removes every registered name regardless of create/readiness/stop state, records the removal outcome, polls every known PID/start-time identity, and forces the overall result to failed if removal or orphan disappearance is not proven.
- Partial JSON is written atomically before execution, at lifecycle transitions/failures, and after cleanup; signal and cleanup outcomes are retained without command stderr, environment values, tokens, or service URLs from process environments.
- Cleanup operations remain bounded: 2s identity capture, 10s forced removal, and 5s identity poll per registered name. Normal readiness/stop bounds remain 30s/20s/5s.

## Re-verification

- Exact corrective scope remains the frozen three paths.
- Probe self/failure-injection tests: 9/9, including create failure, readiness failure, SIGTERM interruption, and cleanup failure.
- `node --test scripts/ci-workflow.test.mjs`: 71/71.
- `git diff --check` and Python AST parse: clean.
- actionlint v1.7.12: clean.
- zizmor 1.26.1: no reportable findings.
- Service image digests, exact slim/full selection, private cgroup-v2/init/read-only topology, metrics/root coherence, graceful/forced exit/OOM/log gates, and forbidden-privilege exclusions remain unchanged and correct.

This is an exact-commit static/security approval only. The PR head is still `f294593a...`, and the currently visible green image/Required CI jobs exercised that older head. Gates 3/4 remain unsatisfied until `b28f5211...` (or a reviewed descendant) is pushed and its hosted slim + full, graceful + forced lifecycle run publishes the digest-bound JSON artifact.

viktor-shcherb:
Guarded exact-head advance evidence

- PR head and remote branch: b28f521100704ec757f8de802b8adac62cc9001a
- Base/main: 168b72b88b12b9e2d25b8414cec92901e007be85
- Approved direct lineage: f294593a2468970a020934885c009f1436bc4f71 -> 03df72ea39b1c2ea81f78d0b85b40bd6564c128c -> b28f521100704ec757f8de802b8adac62cc9001a
- Push: lease-protected fast-forward of the existing fix-crawler/process-tree-sampler-process branch from exact f294593a2468970a020934885c009f1436bc4f71
- Aggregate scope: exactly the original six #8405 files plus the three approved CI-gate files; no #8402 content
- Crawler VERSION: 0.13.638
- Local validation at exact head: 9/9 probe self-tests, 71/71 workflow tests, Ruff, Pyright, actionlint, zizmor, Python byte-compilation, diff checks, and invalid-SHA fail-closed evidence

Hosted checks and the real crawler-image lifecycle artifact are now being monitored to terminal before any further decision. No merge, hold removal, deployment, issue closure, or #8402 action is authorized.

viktor-shcherb:
Hosted lifecycle gate — STOP evidence

Exact-head CI run 33481506821 exercised merge SHA 678054396ec299d943bf0c93c55eff1044562a04 with parents base 168b72b88b12b9e2d25b8414cec92901e007be85 and PR head b28f521100704ec757f8de802b8adac62cc9001a.

- crawler-image job: failed at Exercise real sampler and container shutdown lifecycle
- All preceding steps passed, including exact provenance, both pre-build clean-tree guards, both exact image builds, the probe self-tests, installed-build checks, sync mount contract, and read-only entry-point smoke
- Always-upload step: passed
- Lifecycle artifact: crawler-sampler-container-smoke-b28f521100704ec757f8de802b8adac62cc9001a, artifact ID 9790202172, archive digest sha256:b50194dbd653d9357eb1755ec89c152357aa8ced89124f4dd2dc91e59e4f80a6
- Slim image content ID: sha256:ad66c1cb11ebd3a6d676e9a4f7db006b38ff9ff9f8e2912725080224209b9b9e
- Full image content ID: sha256:9657c2390596596034956b0726c80c49de878c3df74112616e846f14957f9e38
- Both images proved linux/amd64 and exact merge/head/base/source labels

Failure occurred on the first of four required runs: slim graceful exited before live assertions. Both early identity captures failed because the container had already exited. Name-driven forced cleanup succeeded and the artifact reports no surviving known identities, but the known-identity count is zero.

Consequently, only 1/4 runs was attempted. No live topology, cgroup, root compatibility, sampler observation, graceful shutdown, forced shutdown, exit-code, or OOM evidence exists; slim forced and both full-image runs were not attempted. The artifact does not preserve the exited container state or container logs, so it does not identify the underlying startup error.

This is a blocking hosted failure. Advancement stops here. No merge, hold removal, deployment, issue closure, or #8402 action was taken.

Job: https://github.com/colophon-group/jobseek/actions/runs/33481506821/job/99772130434
Artifact: https://github.com/colophon-group/jobseek/actions/runs/33481506821/artifacts/9790202172

viktor-shcherb:
CHANGES REQUIRED — hosted failure diagnosis for exact run `33481506821`, job `99772130434`, artifact `9790202172`.

Exact tested provenance is valid: merge `678054396ec299d943bf0c93c55eff1044562a04`, base `168b72b88b12b9e2d25b8414cec92901e007be85`, PR head `b28f521100704ec757f8de802b8adac62cc9001a`; slim image `sha256:ad66c1cb11ebd3a6d676e9a4f7db006b38ff9ff9f8e2912725080224209b9b9e`; full image `sha256:9657c2390596596034956b0726c80c49de878c3df74112616e846f14957f9e38`.

## What the hosted evidence proves

- The lifecycle step began at `07:22:17.468Z` and reported `container exited before live assertions` at `07:22:17.773Z`: about 0.3 seconds for image launch, failure handling, and diagnosis.
- The first `docker top` attempt failed after 0.036s; the final attempt failed after 0.013s. No host/container PID identity was captured.
- Name-driven forced removal succeeded in 0.016s and no known identity survived, but the known set was empty.
- Only `slim/graceful` was attempted. No sampler child, process tree, cgroup, live metrics, root compatibility, graceful shutdown, forced shutdown, exit code, OOM state, or full-image result was proven.
- PostgreSQL and Redis services were healthy. Both images, provenance guards, builds, self-tests, installed package checks, sync-mount contract, and entry-point help smoke passed.

## Failure classification

This is a **probe image-invocation/configuration failure**, not a production sampler failure.

The slim image default is `uv run --no-sync crawler run`. The probe is the first job step that runs that exact default while also applying `--read-only`. It provides writable `/tmp`, but it does not set `UV_CACHE_DIR` or `UV_NO_CACHE`; uv therefore uses its default `$HOME/.cache/uv` under the now read-only root filesystem. uv documents that every invocation requires a cache directory, even with caching disabled, and its own read-only-Docker issue records that `--no-sync` needs a writable/temporary cache arrangement: [uv cache directory contract](https://docs.astral.sh/uv/concepts/cache/#cache-directory), [astral-sh/uv#12650](https://github.com/astral-sh/uv/issues/12650).

The sub-0.04s disappearance after `docker run` is consistent with the `uv` launcher failing before the Python crawler can remain observable. A database, Redis, or sampler failure would occur after Python startup; the pipeline also retries Redis claim failures and does not perform its first reaper sweep immediately. Production Compose runs the same `uv run --no-sync` command but does **not** make worker/browser roots read-only, so this probe-only mismatch does not identify a production defect.

The exact uv stderr cannot be proven from artifact `9790202172` because the probe deleted the container without preserving logs. The next diagnostic artifact must confirm this classification; if its captured logs show a different cause, stop and reclassify. Do not change sampler or production crawler code based on this run.

## Separate evidence defect

The artifact omitted the cause because `.github/scripts/crawler-sampler-container-smoke.py` currently:

- reads the returned container ID but never stores it;
- validates `state.running` before assigning the inspect record into the case, so an early-exit inspect is discarded;
- retains only status/running/exit/OOM even when inspect succeeds, omitting Docker state error, PID/dead/restarting/paused, started/finished timestamps, and actual path/args;
- reads logs only after a successful planned stop; create/readiness/signal failures reach unconditional removal with no log capture;
- performs final identity capture/removal without a pre-removal diagnostic snapshot.

That means a correctly failing hosted gate can still produce an operationally unusable artifact. This is blocking independently of the uv cache fix.

## Minimal remediation — retain the frozen scope

No production Python, Dockerfile, entrypoint, Compose, migration, VERSION, or docs change is justified. Keep the remediation in the existing gate files:

1. `.github/scripts/crawler-sampler-container-smoke.py`
   - Add `UV_CACHE_DIR=/tmp/uv-cache` to the test-container environment. Keep the exact image entrypoint/CMD, `--read-only`, and existing `/tmp` tmpfs unchanged.
   - Persist the exact returned container ID and name immediately after `docker run`.
   - Store the first inspect result before validation. Include only credential-free diagnostics: image ID, actual path/args, status/running/paused/restarting/dead/PID, exit code, OOM flag, Docker state error, and started/finished timestamps. Never store `Config.Env`.
   - On every create/readiness/signal/stop/cleanup failure, capture inspect plus complete container stdout/stderr **before** forced removal. Reuse this capture on successful stops as well.
   - Apply deterministic redaction before JSON/step-summary persistence: URI userinfo, `Authorization`/Bearer/Basic values, and password/token/secret/key/database/Redis URL assignments must be masked. Record byte/line counts and capture errors; do not silently omit logs.
   - Preserve the existing early/final PID-start identity capture, forced-removal proof, atomic writes, and all lifecycle bounds.

2. `scripts/ci-workflow.test.mjs`
   - Require the writable uv-cache environment while continuing to forbid entrypoint/CMD overrides and privileged/PID-host/host-mount capabilities.
   - Require early-exit diagnostics and redaction fields in the script contract.

No workflow structural change is needed unless implementation requires it for the existing self-test invocation.

## Deterministic tests required

Extend the probe self-tests with an injected container that exits before the first topology snapshot:

- `docker run` returns a fixed 64-hex ID;
- early `docker top` fails;
- inspect reports a fixed non-zero exit, `OOMKilled=false`, state error, PID/status/dead flags, started/finished timestamps, actual path/args, and exact image ID;
- logs contain multiple lines plus seeded URI credentials, Bearer/Basic authorization, and password/token/secret/key/`*_URL` values;
- assert the final partial JSON contains ID/name/full state/path/args, all non-secret log lines, redaction metadata, cleanup outcome, and zero secret literals;
- assert inspect/log capture happens before forced removal;
- inject inspect failure and log-read failure separately and require typed diagnostic errors rather than omission;
- assert `UV_CACHE_DIR=/tmp/uv-cache` is passed to both image runs and no entrypoint/CMD override is introduced.

Re-run probe self/failure tests, `node --test scripts/ci-workflow.test.mjs`, actionlint, zizmor, and `git diff --check`.

## Hosted acceptance

After exact-commit review, the next hosted artifact must either:

- pass all four real runs (slim/full × graceful/forced) with the existing digest-bound lifecycle evidence; or
- if any container exits early, retain enough redacted inspect/log evidence to identify the exact cause before cleanup.

Run `33481506821` remains a valid blocking failure. No merge, hold removal, deployment, issue closure, or downstream advancement is justified.

viktor-shcherb:
CHANGES REQUIRED — independent review of exact candidate `13cdee6bd31c2d032ac399c089d087ba69d8b06c` (parent `b28f521100704ec757f8de802b8adac62cc9001a`).

The remediation architecture is otherwise correct and remains in the frozen two-file gate scope:

- `UV_CACHE_DIR=/tmp/uv-cache` is supplied to both image runs while the exact image entrypoint/CMD remains untouched; `--read-only` and the existing `/tmp` tmpfs remain intact.
- The exact 64-hex ID returned by `docker run`, the selected name, and a sanitized initial inspect record are atomically persisted before runtime validation.
- Initial inspect now retains image ID, actual path/args, and status/running/paused/restarting/dead/PID/exit/OOM/error/start/finish state without retaining `Config.Env`.
- Normal stopped-container removal is preceded by an atomically persisted inspect plus complete redacted stdout/stderr capture. The unconditional outer-finally path independently captures identities and diagnostics before name-driven force removal, including early exit, readiness failure, signal interruption, and typed inspect/log-capture failures.
- Cleanup remains unconditional and fail-closed.

## Blocking defect: escaped quoted assignments can leak credentials

`QUOTED_ASSIGNMENT_PATTERN` terminates at the first quote even when that quote is escaped. An exact standalone call to the candidate redactor demonstrates the leak:

- input: `{"password":"abc\"topsecret"}`
- durable output: `{"password":"[REDACTED]"topsecret"}`

The secret suffix `topsecret` remains in artifact text. This violates the required no-credential-leak property for quoted password/token/secret/key assignments. The current self-test covers ordinary quoted JSON but not escaped delimiters.

Make quoted-assignment redaction escape-aware (or conservatively redact the entire assignment value with a parser/scanner that cannot resume inside the same credential), and add a deterministic test containing unique secret fragments on both sides of an escaped quote. Assert that neither fragment nor the complete secret occurs anywhere in serialized evidence while unrelated lines remain complete.

## Required validation is not clean

The exact candidate passes:

- all 13 probe self/failure tests;
- all 71 workflow tests;
- Pyright: 0 errors;
- actionlint;
- zizmor: no findings;
- `git diff --check`.

Ruff reports three findings on the changed script. Two `SIM117` findings already exist in the parent, but this candidate introduces a new `FLY002` at `.github/scripts/crawler-sampler-container-smoke.py:1953` for the new seeded-log construction. At minimum, remove the new lint regression and rerun the same exact target; do not claim a clean Ruff result while it fails.

## Provenance and acceptance

The candidate has exactly the asserted parent and changes only:

- `.github/scripts/crawler-sampler-container-smoke.py`
- `scripts/ci-workflow.test.mjs`

Workflow provenance/build/lifecycle/upload ordering is unchanged. The live PR head is still `b28f521100704ec757f8de802b8adac62cc9001a`, so `13cdee6b` has not yet received hosted execution. After correcting the redaction leak and reviewing the new exact commit, a hosted rerun is mandatory: all four slim/full × graceful/forced cases must pass, or an early failure must preserve exact redacted inspect/log cause before removal. No merge, deploy, hold removal, issue closure, or downstream advancement is justified yet.

viktor-shcherb:
CHANGES REQUIRED — fresh re-review of exact local candidate `6b6a60793b65bda16631e7ed837e174e2aa98fc9` (parent `13cdee6bd31c2d032ac399c089d087ba69d8b06c`).

The two previously reported corrections are materially improved:

- The escaped JSON case from the prior review is closed. Odd/even backslashes, escaped quote delimiters, multiple assignments, unterminated line-bounded values, Unicode, URI userinfo, and Authorization Bearer/Basic cases are redacted while safe evidence remains.
- The candidate removes the new `FLY002` finding. Targeted Ruff now reports only the same two inherited `SIM117` findings present in `b28f5211`; there is no new lint regression.
- The full prior remediation is unchanged and reconfirmed: writable `/tmp/uv-cache` with exact image entrypoint/CMD, persisted returned ID/name and sanitized inspect before validation, complete redacted stdout/stderr plus inspect captured/persisted before removal, typed capture failures, and unconditional name-driven cleanup.

## Blocking defect: valid POSIX shell value concatenation leaks a suffix

Shell assignment values may concatenate adjacent quoted and unquoted segments without whitespace. `_redact_sensitive_assignments` stops as soon as the initial quoted segment closes, so the rest of the same shell word survives into the durable log artifact.

Exact candidate reproduction through the actual `_redacted_log_stream` artifact path:

- input: `SECRET_KEY='left-fragment-A17Q'right-fragment-Z93M SAFE_LABEL=visible`
- durable text: `SECRET_KEY='[REDACTED]'right-fragment-Z93M SAFE_LABEL=visible`

`right-fragment-Z93M` is still serialized. The canonical apostrophe form, `SECRET_KEY='left'\''right'`, fails for the same reason. This is a real POSIX shell assignment, not an unterminated or malformed input, and it leaves part of the credential visible despite the shell-input requirement.

Continue masking the complete adjacent shell word after a quoted segment (including alternating quoted, backslash-escaped, and unquoted segments) until a line/word or structured-value delimiter is reached. Preserve line bounding and unrelated fields. Add deterministic artifact-path cases for at least:

- `SECRET='left-fragment'right-fragment SAFE_LABEL=visible`;
- `SECRET='left-fragment'\''right-fragment' SAFE_LABEL=visible`;
- alternating `'left'\"middle\"right` segments.

Assert that every unique secret fragment is absent from the serialized `_redacted_log_stream` result and `SAFE_LABEL=visible` remains.

## Exact validation evidence

Passed on `6b6a60793`:

- 14/14 probe self/failure tests;
- 71/71 workflow tests;
- 150/150 complete Workflow Security repository-script tests;
- Pyright: 0 errors;
- actionlint;
- zizmor: no findings;
- `git diff --check`.

Ruff comparison is explicit: `6b6a60793` has only two `SIM117` findings; `b28f5211` has those same two; parent `13cdee6` had those two plus the now-removed `FLY002`.

Scope remains the same two gate files, and workflow provenance/build/lifecycle/upload ordering is unchanged. The live PR head remains `b28f521100704ec757f8de802b8adac62cc9001a`, so this local candidate has not received hosted execution. After the remaining shell leak is corrected and the new exact commit is independently approved, the hosted all-four lifecycle rerun is mandatory. No merge, deployment, hold removal, issue closure, or downstream advancement is justified yet.

viktor-shcherb:
CHANGES REQUIRED — fresh exact re-review of local candidate `3996cf0c92e40834ddba55ed521635b21be2c8ce` (parent `6b6a60793b65bda16631e7ed837e174e2aa98fc9`).

The prior POSIX adjacent-segment blocker is closed for the tested forms: canonical single-quote splice, alternating single/double/unquoted segments, backslash-escaped quotes, escaped spaces, and trailing backslashes are consumed as one sensitive shell word. Structured JSON colon values stop at their unquoted `,`, `;`, `]`, or `}` boundary and preserve following safe fields. URI userinfo and Authorization redaction also remain correct.

## Blocking leak: escaped newline is checked too late

For shell assignments, backslash-newline is a line continuation and therefore part of the same shell word. The lexer sets `escaped = True` on the backslash, but its next loop iteration checks `character in "\r\n"` before the `escaped` branch. It stops at the newline and serializes the continued credential suffix.

Exact reproduction through `_redacted_log_stream` on this candidate:

Input:

```text
PASSWORD=newline-left-H8\
newline-right-I9 SAFE_LABEL=newline-safe
```

Durable text:

```text
PASSWORD=[REDACTED]
newline-right-I9 SAFE_LABEL=newline-safe
```

`newline-right-I9` survives serialized artifact evidence. This directly fails the requested escaped-newline case.

For shell `=` assignments, recognize an unquoted backslash-newline pair as continuation before treating newline as a boundary, and continue masking the same word on the next physical line. Preserve physical line structure in the redacted text if possible so diagnostic line counts remain coherent. Add an artifact-path test with unique fragments before and after the continuation; neither may survive while `SAFE_LABEL=newline-safe` must remain.

## True shell delimiters must preserve safe evidence

The current `structured = operator == ":"` split gives shell `=` assignments no unquoted control-operator delimiters. Therefore:

```text
SECRET=secret-M3;SAFE_LABEL=semicolon-visible;SAFE_MODE=semicolon-enabled
```

becomes only:

```text
SECRET=[REDACTED]
```

No credential leaks in that example, but safe evidence after the true POSIX `;` delimiter is silently discarded. Teach the unquoted shell mode to stop at real unescaped control operators (at minimum `;`, with the lexer treating quoted/escaped occurrences as data), and add a deterministic preservation test. This must not weaken the newly fixed adjacent-segment masking.

## Exact validation evidence

Passed on `3996cf0c92`:

- 15/15 probe self/failure tests;
- 71/71 workflow tests;
- 150/150 complete Workflow Security repository-script tests;
- Pyright: 0 errors;
- actionlint;
- zizmor: no findings;
- `git diff --check`.

Targeted Ruff reports only the same two inherited `SIM117` findings as parent `6b6a6079`; no new lint regression was introduced.

The candidate changes only the same two gate files. The writable uv cache, unmodified entrypoint/CMD, exact ID/name plus initial inspect persistence, redacted inspect/log capture before every removal, typed failure evidence, atomic persistence, unconditional cleanup, provenance guards, workflow ordering, and all-four lifecycle contract remain unchanged and reconfirmed.

The live PR head is still `b28f521100704ec757f8de802b8adac62cc9001a`; this candidate is local and has no hosted execution. After the lexer correction receives exact-commit approval, the hosted slim/full × graceful/forced rerun remains mandatory. No merge, deployment, hold removal, issue closure, or downstream advancement is justified yet.

viktor-shcherb:
APPROVE — fresh exact review of local candidate `958a6f8725aecc1f1828c1a43541dbf4113ae43a` (parent `3996cf0c92e40834ddba55ed521635b21be2c8ce`).

The blockers in the prior review are closed.

## Redaction verification

The assignment lexer now handles a pending escape before testing newline/control boundaries. LF, CRLF, and multiple continued physical lines remain part of the same sensitive shell word; every secret fragment is masked and the original newline structure is preserved in redacted text.

Outside quotes and escapes, the full requested shell control/redirection starts terminate the word and remain in evidence: `;`, `;;`, `;&`, `;;&`, `&`, `&&`, `|`, `||`, `|&`, `(`, `)`, `<`, `>`, `<<`, `>>`, `<<<`, `<>`, `>|`, `<&`, `>&`, `&>`, and `&>>`. The same characters inside quotes or escaped unquoted segments remain part of the masked credential. Following safe fields are preserved.

Structured `:` values remain line-bounded and stop at unquoted `,`, `;`, `]`, or `}`. Nested JSON safe fields remain intact. URI userinfo and Authorization Basic/Bearer redaction are unchanged.

I independently exercised the actual `_redacted_log_stream` serialized-artifact path with:

- LF, CRLF, and multi-continuation values;
- adjacent canonical quote splices and mixed quoted/unquoted segments;
- escaped whitespace, backslashes, quotes, and controls;
- all 22 control/redirection forms above;
- quoted and escaped operators;
- nested structured JSON;
- Redis URI userinfo and Basic authorization.

Result: 41 unique secret fragments, zero survivors; 31 safe markers, zero omissions; all 34 raw physical lines and 33 newline boundaries preserved.

## Exact validation evidence

Passed on `958a6f8725`:

- 16/16 probe self/failure tests;
- 71/71 workflow tests;
- 150/150 complete Workflow Security repository-script tests;
- Pyright: 0 errors;
- actionlint;
- zizmor: no findings;
- `git diff --check`.

Targeted Ruff reports only the same two inherited `SIM117` findings as parent `3996cf0c`; this commit introduces no lint regression.

## Prior remediation and scope

The candidate changes only the same two gate files. The writable `/tmp/uv-cache`, exact unmodified image entrypoint/CMD, returned container ID/name and initial sanitized inspect persistence before validation, complete redacted stdout/stderr plus inspect capture before every removal, typed diagnostic failures, atomic partial evidence, unconditional name-driven cleanup, provenance guards, workflow ordering, and all-four lifecycle contract remain unchanged and reconfirmed.

This is exact-local-commit approval only. The live PR head remains `b28f521100704ec757f8de802b8adac62cc9001a`, so `958a6f8725` has not run in hosted CI. After the reviewed candidate is pushed without change and exact remote provenance is rechecked, the hosted slim/full × graceful/forced lifecycle run is mandatory. Merge, deployment, hold removal, issue closure, and downstream advancement remain blocked until that hosted artifact passes or captures an exact redacted failure cause before cleanup.

viktor-shcherb:
## Guarded rebased-head update

The lease-protected branch update completed from exact prior remote head `b28f521100704ec757f8de802b8adac62cc9001a` to exact rebased head `d0c4ebbb6959b7ba4e7b5833531108b61528da73`.

Immutable identity after the update:

- base: `f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb` (the direct child of the prior reviewed base via #8409)
- head: `d0c4ebbb6959b7ba4e7b5833531108b61528da73`
- head tree: `ca8443f8f5b923711d8549cdfcf3de586b047856`
- aggregate delta: exactly the approved nine paths
- crawler version: `0.13.638`
- all nine approved path blobs are byte-identical to pre-rebase approved candidate `958a6f8725aecc1f1828c1a43541dbf4113ae43a`
- no #8402 transport/routing/contracts surface

The #8409 base delta was disjoint and web/deployment-only. The eight-commit #8407 lineage rebased without conflict.

Local validation on the exact rebased head passed:

- comprehensive focused runtime/metrics/GIL/restart/schema/compat suite: 689/689
- runtime conformance/schema suite: 497/497
- deterministic bindings/protocol/compatibility/privacy/semantics checks
- lifecycle probe self-tests: 16/16
- workflow contract tests: 71/71
- complete Workflow Security repository-script suite: 150/150
- Ruff, format, Pyright, crawler version policy, actionlint, and zizmor
- isolated `jobseek-crawler==0.13.638` wheel build/install, installed sampler/collector imports, and all six installed-role build-info smokes

The full crawler suite was not repeated because the only new base delta is disjoint web/deployment code; its workflow impact is covered by the complete Workflow Security suite.

This update does not authorize merge, deployment, hold removal, issue closure, #8402 work, or downstream burn-in. The remote exact head still requires a formal independent verdict and hosted lifecycle evidence review.

viktor-shcherb (commented):
Verdict: APPROVE — exact rebased head d0c4ebbb6959b7ba4e7b5833531108b61528da73.\n\nGitHub cannot record the APPROVED state because the authenticated identity authored this PR; this COMMENTED review is the formal exact-head review record.\n\nFrozen identity and scope:\n- Base f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb; tree ca8443f8f5b923711d8549cdfcf3de586b047856.\n- Base crawler VERSION 0.13.637; head 0.13.638.\n- Aggregate diff is exactly the nine previously approved #8407 paths, with no #8402 surface.\n- Every changed blob and mode is byte-identical to previously approved head 958a6f8725aecc1f1828c1a43541dbf4113ae43a.\n- The eight rebased commits are linear and single-parent, all stable patch IDs match the old series, diff-check is clean, and there are no merge/conflict artifacts.\n- #8409 contributes exactly ten web/deploy-only paths, disjoint from all nine #8407 paths.\n\nPost-rebase evidence:\n- Focused crawler tests: 145/145.\n- Lifecycle helper self-tests: 16/16.\n- Combined workflow/security assertions: 85/85.\n- Ruff check/format and Pyright: clean.\n- Isolated wheel build/install/import and installed crawler CLI smoke: passed at 0.13.638.\n- Independent audit also ran the full locally runnable workflow-security set: repository Node suites 150/150, observability rollback 12 passed/1 skipped, actionlint clean, zizmor no findings.\n\nA second local full crawler run is intentionally skipped: all crawler artifacts are byte-identical to the already approved head, while the only new-base delta is disjoint web/deploy work; combined-tree workflow/security coverage passed.\n\nThe hosted four-case crawler/sampler lifecycle rerun on this exact SHA remains a mandatory terminal gate before merge. This review does not authorize merge, deploy, hold changes, or baseline capture.

viktor-shcherb:
STOP evidence for exact approved PR head d0c4ebbb6959b7ba4e7b5833531108b61528da73.

Hosted CI run 33487354330 (https://github.com/colophon-group/jobseek/actions/runs/33487354330), job 99790644106 (https://github.com/colophon-group/jobseek/actions/runs/33487354330/job/99790644106), failed in the step named Exercise real sampler and container shutdown lifecycle. The uploaded always-run artifact is ID 9792373005 (crawler-sampler-container-smoke-d0c4...), SHA-256 b1c161197df88c21d6425bde82ac39f51bae89cd11332ea2b6c7b87ce981f4b3.

Provenance is exact: the artifact records base f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb, PR head d0c4ebbb6959b7ba4e7b5833531108b61528da73, and tested merge 6c97707b8729d688143419d86955af05d360a381; the merge ref has those exact base/head parents. Both image labels/content identities were captured, and all preceding image build, provenance, self-test, and installed-image smoke steps passed.

Classification: probe compatibility failure, not an observed crawler lifecycle failure. The container was launched with --stop-timeout 30 and was running with the other required runtime invariants, but the probe read HostConfig.StopTimeout. Docker Engine returns this field as Config.StopTimeout for this inspect shape, so the recorded value was null and the probe raised ProbeError: container stop timeout is not 30 seconds before its live sampler assertions.

Only the slim/graceful scenario began, so the required 4/4 slim/full × graceful/forced lifecycle gate has no acceptance evidence and must remain failed. Cleanup nevertheless succeeded: all four known process identities were gone, no survivors remained, diagnostics were captured, and redaction metadata reported zero exposed credential replacements.

Per the stop rule, no remaining checks were treated as acceptance and no merge/deploy/close action was taken.

viktor-shcherb:
APPROVE — independent exact review of local candidate `edb4745ea1cb46401790fc55375972fcfcbcfa33` (parent `d0c4ebbb6959b7ba4e7b5833531108b61528da73`).

The hosted Docker-inspect compatibility defect is closed without weakening the runtime gate.

## Exact inspect contract

- `_inspect_container` now requires `Config` to be a mapping and reads stop timeout only from `Config.StopTimeout`.
- `HostConfig.StopTimeout` is never read or used as fallback.
- `_validate_container_runtime` requires `type(stop_timeout_seconds) is int` and value exactly `30`; Python/JSON Boolean equality cannot bypass it.
- The hosted-shape fixture accepts integer `30` and rejects missing, `null`, wrong integer `29`, string `"30"`, Boolean `true`, and float `30.0` while planting valid `HostConfig.StopTimeout = 30` bait in every invalid/missing case.
- Existing HostConfig-only invariants (`Init`, cgroup namespace, network, read-only root, and `/tmp` tmpfs) remain unchanged.

## Exact validation evidence

Passed on `edb4745ea1`:

- 17/17 probe self/failure tests, including the new hosted inspect shape;
- 71/71 workflow tests;
- 150/150 complete Workflow Security repository-script tests;
- Pyright: 0 errors;
- actionlint;
- zizmor: no findings;
- `git diff --check`.

Targeted Ruff reports only the same two inherited `SIM117` findings as parent `d0c4ebbb`; this candidate introduces no lint regression.

## Scope and prior semantics

The candidate changes only `.github/scripts/crawler-sampler-container-smoke.py`. The exact image entrypoint/CMD, writable uv cache, returned container ID/name and sanitized inspect persistence before validation, redacted inspect/log capture before every removal, typed failure evidence, atomic partial writes, unconditional name-driven cleanup, lifecycle bounds, all-four matrix, provenance checks, workflow ordering, and hardened redaction lexer are unchanged and reconfirmed.

## STOP evidence audit

Both hosted STOP comments retain the correct run and artifact identities.

For the latest stop:

- run `33487354330`, job `99790644106`, PR head `d0c4ebbb6959b7ba4e7b5833531108b61528da73`;
- tested merge `6c97707b8729d688143419d86955af05d360a381`, with exact parents base `f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb` and head `d0c4ebbb6959b7ba4e7b5833531108b61528da73`;
- artifact ID `9792373005`, name `crawler-sampler-container-smoke-d0c4ebbb6959b7ba4e7b5833531108b61528da73`;
- the STOP comment's `b1c161197df88c21d6425bde82ac39f51bae89cd11332ea2b6c7b87ce981f4b3` is the exact downloaded JSON evidence-file SHA-256;
- GitHub's separate artifact archive/API digest is `sha256:d09b715f0a6da48ca73973f857017c70fbcf75dac87b06dde845c3385416345b`.

The JSON confirms the first slim/graceful case stopped before live assertions with a running container, `stop_timeout_seconds: null`, four known identities, successful diagnostic capture, successful forced removal, zero survivors, and exact image/base/head/merge provenance. The earlier run `33481506821` / artifact `9790202172` also matches its recorded archive digest `sha256:b50194dbd653d9357eb1755ec89c152357aa8ced89124f4dd2dc91e59e4f80a6`.

This is exact-local-commit approval only. The live PR head remains `d0c4ebbb6959b7ba4e7b5833531108b61528da73`; `edb4745ea1` has not received hosted execution. After it is pushed unchanged and exact remote provenance is rechecked, a fresh hosted slim/full × graceful/forced 4/4 run is mandatory. Merge, deployment, hold removal, issue closure, and downstream advancement remain blocked until that artifact passes or records an exact redacted failure cause before cleanup.

viktor-shcherb:
Exact-head advance evidence for PR #8407.

The branch advanced by guarded fast-forward from d0c4ebbb6959b7ba4e7b5833531108b61528da73 to independently approved edb4745ea1cb46401790fc55375972fcfcbcfa33. The new commit is the single direct child of d0c4; its tree is 39c0bd825aa6de7fb59444e45072fe2937c10644. A fresh remote verification after the lease-protected push confirms PR base f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb and exact PR head edb4745ea1cb46401790fc55375972fcfcbcfa33.

Aggregate scope remains exactly nine paths:
- .github/scripts/crawler-sampler-container-smoke.py
- .github/workflows/ci.yml
- apps/crawler/VERSION
- apps/crawler/src/metrics.py
- apps/crawler/src/runtime_cost/process_tree.py
- apps/crawler/tests/test_metrics_server.py
- apps/crawler/tests/test_runtime_process_tree.py
- docs/23-go-lightpanda-migration.md
- scripts/ci-workflow.test.mjs

Crawler version remains exactly 0.13.638, and the aggregate commits/diff contain no #8402 content.

The sole new commit is probe/static-test scope. It reads Docker inspect Config.StopTimeout, requires an exact integer value of 30, and never accepts null or falls back to unrelated HostConfig.StopTimeout. Its hosted-shape self-test accepts Config.StopTimeout=30 while explicitly rejecting missing, null, 29, string, boolean, and float values even when HostConfig.StopTimeout is set to 30.

Pre-push verification was green: probe self-test 17/17; CI workflow tests 71/71; repository security tests 150/150 plus rollback tests; crawler Ruff check/format and Pyright; standalone probe Pyright/compile and scoped Ruff; actionlint; zizmor; clean diff/worktree.

Hosted exact-head checks and the required 4/4 slim/full × graceful/forced lifecycle artifact remain pending. No merge, hold, deployment, close, or #8402 action is authorized by this advance.

viktor-shcherb:
STOP evidence for exact PR #8407 head edb4745ea1cb46401790fc55375972fcfcbcfa33.

The exact-head Crawler Deploy Gate context became terminal FAILURE at 2026-09-01T08:45:44Z with description: Crawler deployment is held. Target run: https://github.com/colophon-group/jobseek/actions/runs/33488676745

The trusted pull_request_target reconciliation workflow itself completed successfully at the exact head. Its log classified PR #8407 as changing crawler runtime paths while deployment hold issue #8228 is active:
- apps/crawler/VERSION
- apps/crawler/src/metrics.py
- apps/crawler/src/runtime_cost/process_tree.py
- apps/crawler/tests/test_metrics_server.py
- apps/crawler/tests/test_runtime_process_tree.py

Classification: expected policy/operations gate failure caused by the pre-existing deployment hold, not a sampler implementation failure and not lifecycle acceptance evidence. The status API confirms SHA edb4745ea1cb46401790fc55375972fcfcbcfa33, context Crawler Deploy Gate, state failure, and the same target run.

Per the explicit failure stop rule, pending exact-head checks are not being treated as acceptance and are no longer being monitored. No lifecycle artifact was retrieved or accepted; the required 4/4 slim/full × graceful/forced matrix and its topology, cgroup, root, sample, shutdown, OOM, orphan, provenance, and redaction assertions therefore remain unproven for this head.

No merge, hold mutation, deployment, close, or #8402 action was taken.

viktor-shcherb:
Exact-head hosted verification and lifecycle audit for PR #8407 head edb4745ea1cb46401790fc55375972fcfcbcfa33.

Ordinary hosted verdict: PASS. All non-deploy-gate checks are terminal with 34 passing checks, including Required CI, and one expected Probe new/changed boards skip. CI run 33488679882 completed successfully at the exact head: https://github.com/colophon-group/jobseek/actions/runs/33488679882
CodeQL run 33488680005 and Crawler Runtime v1 Contracts run 33488679823 also completed successfully.

Crawler image lifecycle job 99794910633 passed, including both Exercise real sampler and container shutdown lifecycle and the always-run artifact upload:
https://github.com/colophon-group/jobseek/actions/runs/33488679882/job/99794910633

Audited artifact:
- ID: 9792917567
- Name: crawler-sampler-container-smoke-edb4745ea1cb46401790fc55375972fcfcbcfa33
- JSON SHA-256: 12d7dbf3dec8f20ee35fbdfe4cb107b799c642ad027ed795fa9fb2b051e3dada
- Overall status: passed
- Base: f2e1da9f91950c73aa98cb592fbe6b05feb6fdbb
- PR head: edb4745ea1cb46401790fc55375972fcfcbcfa33
- Tested merge: 8e7da1002c82cef6f8d47c03dad42deb23ec62c5
- Tested merge parents: exact base then exact PR head
- Tested merge tree: 39c0bd825aa6de7fb59444e45072fe2937c10644

Image provenance is exact for both linux/amd64 images. Slim content ID is sha256:35331990c71e34f4cfe0c0cc65749c1aa4f43600a1148f7035ec307784cf25cb; full content ID is sha256:00a4437ddc2671bb239fd73c2f7d50e1a7788b2038689c9170ace6d53121b7fb. Source, revision, base, and PR-head labels all match the exact tested merge and PR.

Required 4/4 matrix independently audited:
- slim/graceful: passed; 1 sample; 2 descendants; exit 0; stop 0.193s; 6 known identities cleaned
- slim/forced: passed; 1 sample; 2 descendants; exit 137; stop 1.071s; 6 known identities cleaned
- full/graceful: passed; 1 sample; 3 descendants; exit 0; stop 0.194s; 8 known identities cleaned
- full/forced: passed; 1 sample; 3 descendants; exit 137; stop 1.068s; 8 known identities cleaned

All four cases proved Docker init, private cgroup namespace, host networking, read-only root, bounded hardened tmpfs, and exact integer Config.StopTimeout=30. Topology was exact: PID 1 to uv launcher to crawler to sampler, with Xvfb as the required launcher child only in full-image cases; every required process shared cgroup v2 path 0::/.

Every case recorded one successful coherent 0.5s sampler observation, zero failures, zero gaps and gap reasons, one sampler start, positive root/tree CPU and RSS with tree totals at least root totals, expected descendant count, and root /proc compatibility within recorded CPU, RSS, and sample-age tolerances.

Graceful cases exited 0 with pipeline.stopped, cli.shutting_down, and cli.stopped markers. Forced cases SIGSTOPed the exact crawler identity, exited 137 within the strict bound, and did not falsely record cli.stopped. Every case reported OOM false, zero surviving shutdown identities, successful name-driven final cleanup, and zero final survivors.

All inspect/log diagnostic redaction counters were zero, byte counts remained coherent, and an independent credential-pattern scan of captured streams was clean.

Qualified verdict: the implementation and required 4/4 lifecycle gate PASS at exact head edb4745ea1cb46401790fc55375972fcfcbcfa33. The separate Crawler Deploy Gate remains its previously classified expected governance failure because hold #8228 is active; that hold was not mutated and is not a lifecycle defect.

No merge, deployment, close, hold mutation, or #8402 action was taken.

PR #8460, opened 2026-09-07
Fix monitor description ownership

viktor-shcherb:
Closes #8458
Part of #8453

## Summary

- keep the monitor teaser as a first-seen availability fallback
- make detail enrichment authoritative for existing description locales
- use insert-only fallback semantics for genuinely new monitor locales, preventing locale metadata from pointing at a missing R2 key
- preserve monitor ownership for boards that do not enrich descriptions
- cover the full teaser → failed detail → full detail → later monitor lifecycle
- repair pre-existing required-CI rot exposed on September 7: bounded pytest display IDs, a fixed clock for an expired Swiss fixture, and the current Lightpanda census for already-merged board data

This stops monitor teasers and scraper full bodies from alternating in R2 while retaining failure-safe content for new postings.

## Verification

- `uv run pytest tests/test_batch.py tests/test_rich_monitor_scheduling.py tests/test_enrich_description.py`: 331 passed
- CI-rot regression set: 9 passed
- `uv run ruff check` and `uv run ruff format --check` on changed crawler files
- independent critic re-review: prior locale/R2-key blocker resolved; no remaining merge blocker

## Operational note

For affected boards with `rescrape_policy=never`, the first successful full detail body intentionally remains authoritative over later monitor teasers. Later upstream full-body edits require periodic rescraping or a bounded manual re-enrich. Post-merge repair will start with DuPont under an explicit write budget.

No alert implementation is included.

Issue #8453, opened 2026-09-07
[Epic] Reduce Cloudflare R2 usage and verify cost controls


viktor-shcherb:
## Outcome

Remove the company OG write-amplification path, eliminate avoidable job-description rewrites, and bound future Cloudflare R2 cost without compromising cache correctness, description availability, or historical saved-job rendering.

The August 2026 investigation found 3.708M successful PUTs versus 1.273M in July. Company OG writes accounted for 1.868M (76.7%) of the increase. Storage grew separately from 3.02M to 4.65M objects during August, but its modeled month-over-month charge increase was only about $0.09.

## Decisions

- No alert implementation in this epic.
- Do not merge a workflow-only removal of `--force`: the current non-force path can publish a fresh completion marker over stale stable keys.
- Ship source-aware incremental OG planning, marker correctness, and fail-closed write budgets together.
- Treat Cloudflare GraphQL counts as operational analytics, not the billing ledger.
- No Glacier.
- Do not deploy an age-based lifecycle rule for `job/**`; active descriptions can be old and keys do not encode posting state.
- Do not move job descriptions to R2 Infrequent Access at the current scale. It remains an online/direct-access option for later, but current modeled savings are negligible and bulk transition writes would overwhelm the benefit.
- Treat saved/applied inactive postings as protected online product data. Both the historical Typesense document and R2 description must remain renderable.
- Before any future tiering or deletion, explicitly decide whether the retention hold means currently referenced or ever saved at least once; fail closed when hold data is unavailable.
- Defer destructive storage cleanup and bucket migration until #8456 proves safety and value.

### Immediate cost and correctness fixes

- [x] #8454 — revision-based incremental company OG prewarming and write budgets
- [x] #8455 — byte-authoritative job-description upload deduplication
- [x] #8458 — prevent monitor teasers overwriting enriched descriptions

### Measured follow-up

- [ ] #8456 — explain the `job/**` object gap and define saved/history retention; IA is deferred pending positive economics (low priority)
- [ ] #8457 — decide producer credential/bucket isolation after residual usage is measured

### Governance

- [ ] #8478 — require independent approval for manual full rebuilds

## Completion

- Automatic company-data changes cannot trigger a full R2 rewrite.
- A one-company update rewrites only that company's locale cards plus bounded markers.
- Description metadata churn cannot PUT unchanged HTML.
- Monitor teasers cannot replace enriched full descriptions.
[...]

Issue #8458, opened 2026-09-07
Stop monitor teasers overwriting enriched R2 descriptions


viktor-shcherb:
Parent: #8453
Related: #8455

## Problem

Boards can expose a partial monitor description while configuring the detail scraper to enrich `description`. For existing/touched rows, `processing/board.py` still stages the monitor teaser, then the scraper stages the full body. This creates alternating R2 PUTs and can leave users with the teaser.

Live read-only evidence for `dupont-careers-phenom` on 2026-09-07:

- 618 description rows total; all marked uploaded.
- 600 rows have hashes matching the HTML-only monitor path and median body size 351 bytes.
- 18 rows have scraper composite hashes and median body size 3,826 bytes.
- Thirteen configs combine monitor `fields.description` with scraper description enrichment, including DuPont, eBay, Honda, Kaiser, Merck/MSD, PwC US, Southwest, Toyota, US Bank, FAB, Romande Energie, and TikTok.

## Outcome

Make the detail scraper authoritative for description updates once a row exists whenever `enrich` includes `description`, while preserving a monitor teaser as a first-seen availability fallback until the initial detail scrape succeeds.

## Scope

- For existing/touched postings on boards whose effective enrich fields include `description`, do not stage or UPSERT the monitor description into `descriptions`.
- Keep initial insert teaser behavior for availability and failure safety.
- Ensure the detail scrape replaces the teaser once and thereafter owns description changes.
- Add tests for first insert, touched existing row, successful enrich, enrich failure, locale behavior, and boards without description enrichment.
- After deployment, run a controlled re-enrich/backfill for affected rows; start with DuPont and cap expected R2 PUTs.

## Acceptance

- A monitor refresh cannot replace an existing full description with a teaser.
- A new posting remains available with its teaser until detail enrichment completes.
- Successful detail enrichment produces one replacement PUT, not an alternating daily pair.
- The affected-board inventory and backfill write budget are logged before execution.

## Post-merge recon

Verify DuPont full-body size/checksum samples, absence of teaser regressions across a complete monitor + scrape cycle, bounded backfill PUTs, and no increase in postings lacking content.

No alert implementation is included.

viktor-shcherb:
Independent design review added one required guardrail before merge:

- Existing full descriptions must not be overwritten in the same locale.
[...]

PR #8465, opened 2026-09-07
Deduplicate R2 description uploads by HTML

### LINKED ISSUE #8455: Deduplicate R2 job-description uploads against stored bytes
Parent: #8453

## Problem

`job/**` successful PUTs increased from 1.260M in July to 1.827M in August. The crawler hashes HTML plus title, locations, extras, metadata, dates, salary, and job-type fields, while `put_description()` uploads only HTML. Hash-only metadata churn can therefore schedule a rewrite of identical object bytes. Redis work also carries a stale scalar posting hash across recurring reschedules, while descriptions are locale-keyed.

## Outcome

Make the `(posting_id, locale)` description row authoritative and schedule R2 only when the exact stored HTML bytes change, without a one-time migration rewrite or broken in-flight claims. Keep the design compatible with #7945.

## Scope

- Compare incoming normalized HTML with the existing `descriptions.html` in the atomic UPSERT.
- If HTML is equal, preserve hash, upload/retry state, and in-flight claim identity even when metadata differs.
- If HTML differs, store the HTML body hash and schedule exactly one upload.
- Remove unsafe early-skip dependence on scalar `job_posting.description_r2_hash`; retain the database final guard.
- Preserve superseded-write behavior, locale changes, retries, exporter semantics, and public object keys.
- Add bounded diagnostic evidence for board, old/new HTML checksum, and whether upload state changed; begin validation with `dupont-careers-phenom`. Do not add high-cardinality Prometheus labels.

## Acceptance

- Same HTML plus changed metadata produces no pending R2 upload.
- Changed HTML produces exactly one pending upload.
- Existing composite hashes do not cause a migration rewrite wave.
- An equal-HTML UPSERT during an in-flight drain preserves the claimed hash so completion still succeeds.
- Locale, superseded, retry, and exporter tests remain green.

## Post-merge recon

Observe at least one full scrape interval. Compare PUTs per completed scrape to a matched pre-deploy period, verify repeated DuPont keys stop rewriting when HTML is unchanged, check pending/claimed rows, and sample public descriptions/checksums.

No alert implementation is included.

viktor-shcherb:
Closes #8455
Part of #8453

## Summary

- make the exact normalized HTML bytes stored in R2 the authoritative description content version
- lock and compare the existing `(posting_id, locale)` row atomically before scheduling an upload
- preserve legacy hashes, completed state, retry state, and in-flight claims when HTML is unchanged
- reset one locale to a new byte hash and pending state only when its HTML actually changes
- add sampled checksum/state diagnostics without logging description content
- document the byte-authoritative contract and migration behavior

This removes metadata-only R2 PUTs without a mass hash migration or availability gap. It composes with #8460: monitor-owned teaser fallbacks remain insert-only, while a successful detail scrape remains authoritative for existing locales.

## Verification

- post-rebase combined ownership/staging/drain/enrichment suite: 369 passed
- independent critic: approved PostgreSQL locking and drain semantics after real PostgreSQL 18 concurrency checks
- additional critic suites: 281 staging/drain/batch + 227 exporter/pipeline/Redis tests
- Ruff check/format and `git diff --check`

## Operational recon

After deployment, compare sampled `r2.description_dedupe_sample` events and description-row lock latency with Cloudflare `job/**` successful PUTs. Existing composite hashes are intentionally preserved until HTML genuinely changes; that avoids a migration rewrite wave.

No alert implementation is included.


viktor-shcherb:
Independent critic re-review: **APPROVE / merge**.

Verified the rebased commit preserves #8458's monitor/detail ownership split, keeps equal-HTML and in-flight R2 state intact, resolves the combined lifecycle mock correctly, and advances crawler `VERSION` to `0.13.641`. Independent focused run: 289 tests passed; Ruff and diff checks clean. Earlier PostgreSQL 18 checks also covered equal/changed content, legacy hashes, in-flight claims, first-insert concurrency, and locale concurrency.

PR #8811, opened 2026-09-10
Trigger Murmur deploy for shared schema inputs

viktor-shcherb:
## Summary
- trigger the Murmur image workflow when either selectively copied web schema input changes
- add an exact, push-block-scoped regression test for both paths

## Why
The Murmur Dockerfile compiles `apps/web/src/db/schema.ts` and `apps/web/src/lib/notifications/contracts.ts`, but a change limited to those files previously did not trigger the image deployment. That allowed build-context drift to remain latent until an unrelated coupled crawler release.

## Verification
- `node --test scripts/ci-workflow.test.mjs`: 85 passed
- actionlint: passed
- zizmor: no findings
- `git diff --check`: passed
- independent critic approved exact commit `a775218d3234aec8fa78492764f3fbb32fa0d725`

No rendered UI and no crawler runtime change.

Follow-up to #8807.

### DIFF
--- a/.github/workflows/deploy-murmur-shim.yml
+++ b/.github/workflows/deploy-murmur-shim.yml
@@ -7,9 +7,9 @@
 #
 # Trigger model:
 #   - `push` to `murmur-demo` touching the shim source, the crawler
-#     compose (where the murmur-shim service is declared by H3), or
-#     this workflow file. The path filter prevents accidental redeploys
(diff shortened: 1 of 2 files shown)


viktor-shcherb:
Closing without merge: the owner has paused Murmur integration work. The reviewed commit remains on the branch if this hardening is resumed later.

PR #8851, opened 2026-09-11
Wait for stable Lightpanda host identity

viktor-shcherb:
Phase A passed the reboot itself but compared the protected-container snapshot immediately after Docker became reachable. A controlled follow-up reboot showed the raw protected Docker inspections are byte-for-byte stable once daemon metadata settles.

Retry the complete exact empty-host snapshot plus receipt validation for a bounded window. Any persistent component, container, or protected-service drift still fails closed; only a fully exact match publishes the receipt.

Tracks #8738 and follows #8850.


viktor-shcherb:
Closing unmerged after adversarial review. The proposed broad retry could mask temporary safety violations, did not enforce its claimed wall-clock bound, and could leave a remotely published receipt after a failed workflow. The unchanged strict Phase A subsequently passed in run 34592263248 after Docker completed its one-time first-reboot bookkeeping transition, so this relaxation is unnecessary. No code from this PR will ship.

PR #9169, opened 2026-09-14
Fix worker proxy routing

### LINKED ISSUE #9167: [daily-error-review] Redis scrape workers bypass configured proxy (PracticeMatch backlog)
## Summary

The 2026-09-14 daily error report retained 12 terminal `ConnectTimeout` scrape failures for `www.practicematch.com` in a 33-second burst. Investigation found a generic worker-path defect: Redis scrape workers load each posting's scraper metadata but do not construct a proxy-aware HTTP client when `scraper_config.proxy=true`.

All three configured PracticeMatch boards require the proxy. A read-only production probe against the same detail shape timed out directly after 15.78 seconds but returned HTTP 200 and 83,616 bytes through the configured proxy in 2.50 seconds.

## Impact

Production currently has 941 active PracticeMatch postings; all 941 lack descriptions and 493 are stalled (`scrape_failures >= 3`, `next_scrape_at IS NULL`). Ten of the 12 retained report URLs are in that stalled/missing cohort and two remain scheduled/missing.

The shared host circuit introduced for #5707 behaved correctly: it opened on the third timeout and deferred 203 later siblings. Twelve requests were already in flight because work was admitted every ~2 seconds while the direct connect timeout was ~30 seconds. Redis maxmemory pressure then caused some circuit-update/reschedule failures; that independent dependency is resolved by #9129.

## Root cause

The Redis worker path passes its shared direct client into scraping even when the loaded scraper config requests a proxy. The legacy batch path already constructs a proxy-aware client, which is why workspace/live audits did not expose the production-path mismatch.

## Durable resolution

- Make Redis scrape workers select and own a configured client when either `proxy` or `ssl_verify` overrides are requested.
- Reuse the shared client when no override is needed and always close only worker-owned clients.
- Add worker-boundary tests for proxy-only, SSL-only/combined, and default reuse.
- After deployment, recover the exact stalled PracticeMatch cohort and verify descriptions progress without new direct `ConnectTimeout` failures.

viktor-shcherb:
## Summary

- honor per-board proxy and TLS transport settings in Redis scrape workers
- close worker-owned clients on every path while retaining the shared default client
- add exact-board scoping to stalled-scrape recovery so affected cohorts can be replayed safely
- document the scoped dry-run workflow

## Why

PracticeMatch detail pages require the configured proxy. The Redis worker loaded `scraper_config.proxy=true` but passed its shared direct client, producing 12 terminal `ConnectTimeout` failures and leaving 493 active postings stalled without descriptions. The batch path already honored this setting.

A global immediate recovery would release more than 31,000 stalled rows, so this PR also makes operational recovery explicitly board-scoped and dry-runnable.

## Verification

- `uv run ruff format --check src tests`
- `uv run ruff check src tests`
- `uv run pyright src/workers/pipeline.py src/retry_stalled.py`
- `uv run pytest tests/test_pipeline.py tests/test_retry_stalled.py tests/test_http.py tests/test_batch.py -q` (386 passed)
- `uv run pytest tests/ -q` (12,586 passed, 113 skipped)

Closes #9167


viktor-shcherb:
Independent subagent critic verdict for exact head `957059f6237ae175beb3f92418332d541a2cf667`: **APPROVE**.

No blocking findings. The critic verified proxy-only, TLS-only, combined, and `skip_ssl + proxy` routing; owned-client cleanup; fail-closed proxy misconfiguration; exact parameterized board scoping with dry-run/update predicate parity; Ruff, Pyright, and 386 focused tests.

PR #9170, opened 2026-09-14
Optimize scoped scrape recovery

### LINKED ISSUE #9167: [daily-error-review] Redis scrape workers bypass configured proxy (PracticeMatch backlog)

viktor-shcherb:
## Summary

- split global and board-scoped stalled-scrape recovery queries
- materialize the exact board IDs before joining through the existing active-postings-by-board index
- preserve identical stalled-row predicates for dry-run and update paths
- ensure an explicit empty board list remains a zero-row scope

## Production evidence

After #9169 deployed, the exact three-board dry-run found 499 candidates but the matching update was canceled atomically by the 30-second PostgreSQL statement timeout. No postings were released.

`EXPLAIN` showed the nullable-scope `OR` selected a full parallel scan of `job_posting`. The replacement query uses `job_board_board_slug_key` followed by `idx_jp_board_active_last_seen`. Read-only production `EXPLAIN ANALYZE` returned the same 499 targets in about 3 ms.

This does not raise the statement timeout, add a broad index, or perform manual recovery.

## Verification

- `uv run ruff format --check src tests`
- `uv run ruff check src tests`
- `uv run pyright src/retry_stalled.py`
- `uv run pytest tests/test_retry_stalled.py -q` (17 passed)
- `uv run pytest tests/ -q` (12,589 passed, 113 skipped)

Follow-up to #9169. Closes #9167


viktor-shcherb:
Critic review of prior head `809e7c799723339e8807903ac13fe57747d7a25b` correctly **requested changes**: `--max-age-days 0` could reselect a row that restalled during the same multi-batch invocation because each statement used a later `now()`.

Addressed in new head `4df4b2cd5e77af0a8bc08f9614c023e56e4d52ee` by capturing one UTC cutoff per invocation and passing it to every count/update batch. Added a regression asserting one clock read and the same cutoff across all batches. Final production read-only plan check returned the same 499 targets via the intended indexes in ~196 ms cold. Full suite: 12,590 passed, 113 skipped. Fresh exact-SHA critic re-review and CI are pending.

viktor-shcherb:
Independent critic re-review for exact head `4df4b2cd5e77af0a8bc08f9614c023e56e4d52ee`: **APPROVE**.

No blocking findings. The critic verified the fixed-cutoff re-entry fix, scoped/unscoped parameter numbering and predicate parity, empty-list fail-closed behavior, 18 focused tests, and the production indexed plan evidence. Non-blocking note: a database-sourced cutoff could eliminate theoretical inter-host clock skew; synchronized production clocks plus enqueue/request latency make that immaterial for this bounded recovery.

viktor-shcherb:
Independent critic re-review: APPROVED exact head `41005d107526ab5ce4d484c203590cb8267735e2`. Verified parent is current base `e178ce02ef4c82e7e434c27af9be1ed9a85446d6`; retry source/tests are byte-identical to the previously approved implementation, the rebase delta is only `VERSION` to `0.13.744`. Fixed-cutoff termination, exact-board scope safety, SQL parameter numbering, predicate parity, 18 focused tests, Ruff, and Pyright all pass. No blockers.

PR #9194, opened 2026-09-15
Fix European Aquatics job boundaries

### LINKED ISSUE #7971: [daily-error-review] workers.european_aquatics_inline_guard (novel 8 observed; one board)
## Summary

HTTP workers emitted eight visible terminal monitor errors for one European Aquatics board during the explicit 2026-08-24 09:00 UTC -> 2026-08-25 09:00 UTC window. Current-generation application logs begin at 18:23 UTC, so this is an observed-slice signal rather than a full-window trend. Five runs found no accepted jobs without the configured explicit empty state; after the configuration changed, three later runs rejected the alternate representation because it omitted required text. This closely follows #7905 and #7909.

## Signal

The unit is one terminal structured `batch.monitor.error` for this board, combining two linked fail-closed stems from the same configuration lifecycle.

| day | count |
|---|---:|
| 2026-08-21 (-4) | 0 in retained report evidence |
| 2026-08-22 (-3) | 0 in retained report evidence |
| 2026-08-23 (-2) | 0 in retained report evidence |
| 2026-08-24 (-1) | unavailable — no retained report |
| 2026-08-25 (today) | **8 observed: 5 explicit-empty mismatches + 3 required-text omissions** |

First seen: **2026-08-24 18:23:52 UTC**. Last seen: **2026-08-25 05:04:56 UTC**.

## Sample log line

```json
{"event":"batch.monitor.error","crawler_type":"inline","error":"inline fetch from https://<redacted-host>/<redacted-path> omitted required text","level":"error","timestamp":"2026-08-25T05:04:56Z"}
```

## Root cause hypothesis

1. `apps/crawler/data/boards.csv` may rely on a `fetch_contains` marker that is not stable in the selected read-only representation. Re-verify the current response shape and bind the guard to a durable, visible marker.
2. `apps/crawler/src/core/monitors/inline.py` may correctly reject a hidden or absent empty-state marker after all historical roles are excluded. Update the scoped empty-state/extraction contract with live-derived reduced fixtures while keeping the fail-closed behavior.

## Impact

One posting inventory can remain stale until the source representation or configuration is corrected. The explicit failure protects existing postings from an unsafe empty result; no broader worker outage, restart, or OOM is verified.

viktor-shcherb:
## Summary

- bind European Aquatics job items to Elementor heading-widget titles instead of every semantic `h2`
- keep nested job sections in one full HTML description and parse both current and legacy location labels
- add a reduced live regression fixture, lock the monitor contract, and refresh the capability census

## Root cause

The monitor was operationally green but used a tag-only `h2` item boundary. The current vacancy body contains five semantic `h2` section headings, so one official vacancy became six active postings: the real role plus Job Purpose, Key Responsibilities, Qualifications & Experience, Additional, and Salary & Details.

## Durable prevention

- the boundary and title step now require `class=elementor-heading-title`
- description capture explicitly extends to the bounded item end, preserving nested headings
- regression coverage recreates the live nested-heading layout and requires exactly one job with its complete description
- contract assertions and the generated capability census make selector drift visible in CI

## Verification

- 219 focused inline/census/CSV tests passed
- full crawler suite: 12,661 passed, 113 skipped; four provider-interface tests initially lacked optional SDKs, then passed after `uv sync --extra enrich`
- Ruff, Pyright, `ws validate`, format hooks, and diff checks passed
- live updated configuration discovers exactly one job, location Belgrade, and the complete 7,323-character description

Closes #7971


viktor-shcherb:
Independent read-only critic verdict:

> APPROVE exact SHA `39d469b98f3a31f0cc22586cfa35d695e7588564`

No actionable findings. The critic independently reproduced 6 postings on the base versus exactly 1 on the head, verified the 7,323-character description and Belgrade location, confirmed the real posting retains its source identity while only phantom URLs enter normal missing lifecycle, and passed the 219 focused regression/census/CSV tests.

Issue #7741, opened 2026-08-21
[daily-error-review] browser.njoyn_pagination_no_jobs (novel 1; one board)


viktor-shcherb:
## Summary

Browser-1 failed one Njoyn board monitor because submitting the next page added no unique jobs while the advertised inventory was incomplete. The explicit window is **2026-08-20 09:08 UTC -> 2026-08-21 09:08 UTC**; current-generation logs cover only 08:02-09:08 UTC, so this is a contained observed-slice signal rather than a full-window trend. The posting set may remain stale, but the fail-closed guard prevented acceptance of a partial inventory.

## Signal

The unit is one terminal browser-1 `batch.monitor.error` with `crawler_type=njoyn` and stable message `Njoyn pagination added no jobs after page <n>; collected <n> of <n> jobs`.

| day | count |
|---|---:|
| 2026-08-17 (-4) | unavailable — no retained report |
| 2026-08-18 (-3) | unavailable — no retained report |
| 2026-08-19 (-2) | unavailable — no retained report |
| 2026-08-20 (-1) | 0 in the full-window error extract |
| 2026-08-21 (today) | 1 in the retained 08:02-09:08 UTC slice |

First and last seen: **2026-08-21 08:31:18 UTC**. The class is absent from every retained prior report and targeted non-company-request issue search.

## Sample log line

```json
{"event":"batch.monitor.error","crawler_type":"njoyn","error":"Njoyn pagination added no jobs after page <n>; collected <n> of <n> jobs","level":"error","timestamp":"2026-08-21T08:31:18Z"}
```

## Root cause hypothesis

1. `apps/crawler/src/core/monitors/njoyn.py` may treat a repeated provider page as a successful page transition before comparing unique URLs; retain a stable page fingerprint and retry the transition before failing.
2. `apps/crawler/src/core/monitors/njoyn.py` may encounter live inventory churn or duplicate listings while using the advertised total as a fixed snapshot; add a bounded consistency pass without weakening the completeness guard.

## Impact

One Njoyn posting-set refresh failed and may remain stale until retry. Browser-1 continued processing; no service restart, OOM, or broader Njoyn outage is verified.

viktor-shcherb:
Severity update from the 2026-09-14/15 production review: **high**.

The original single contained failure developed into a repeated production outage for the only Njoyn board:

- 8 terminal `pagination added no jobs` failures were retained between 2026-09-02 and 2026-09-14.
[...]

PR #9271, opened 2026-09-15
Recover blocked rendered job details

viktor-shcherb:
## Summary
- validate rendered JSON-LD content while its proxy context is still open, so HTTP-200 challenge shells receive origin-scoped failure accounting
- add bounded proxy-context rotation and an explicit direct fallback that is reachable only after a typed origin block
- sanitize Njoyn challenge redirects by selecting the provider guard from the requested origin, and extend configuration census, tests, docs, and daily-review guidance

## Production evidence
The 0.13.765 acceptance check separated extraction from transport: the same active CGI detail page returned Njoyn's small XWP rejection shell through all five configured proxy slots, while the production worker's headful direct path extracted title, description, location, employment type, and posting date. The persisted board remains active with zero monitor failures, so this is a detail transport recovery gap rather than inventory or JSON-LD drift.

## Safety invariants
- maximum five proxy attempts
- direct egress requires explicit board opt-in plus a typed origin block observed by the current call
- pool unavailability alone never authorizes direct fallback
- content checks run before context exit so the selected origin/slot can be quarantined
- Njoyn errors never include response bodies or opaque challenge redirect URLs

## Verification
- 12,754 passed, 113 skipped
- 379 focused browser/JSON-LD/Njoyn/census tests passed
- Ruff and Pyright clean
- Lightpanda census check: 297 records, 496 browser boards
- workspace CSV validation passed
- production direct-path read-only probe extracted all required structured fields

Related to #7741. The incident remains open until this is deployed and scheduler-owned retries converge.


viktor-shcherb:
Independent exact-SHA critic: **APPROVE — no actionable findings** for `4bc83652e6358ef79360a75acb7e14f43cd44145`.

The critic verified SHA equality, in-context slot attribution, requested-vs-final URL sanitization, five-attempt cap, explicit typed-block-only direct fallback, fail-closed initial pool exhaustion, cancellation propagation, BROWSER_KEYS/census compatibility, and runbook coverage. Independent gates: 379 focused tests; Ruff, Pyright, census, and diff checks clean.

PR #9275, opened 2026-09-15
Open host circuit for rendered browser challenges

viktor-shcherb:
## Summary
- promote typed browser-rendered bot challenges into the existing shared per-origin scrape circuit breaker
- attribute challenge redirects to the stable requested job origin and let a later validated render supersede an earlier blocked attempt
- add end-to-end circuit tests proving three blocked runs open the circuit and persist the learned scrape host

## Production impact
Njoyn's refreshed 10-slot pool remains uniformly rejected by the origin. Without this bridge, browser navigation bypasses the tracked httpx transport, so 3,000+ posting retries can each exhaust proxy rotation and direct fallback independently. This change makes the existing 3-failure/30-minute host circuit protect the whole detail cohort while retaining one half-open recovery probe. It does not solve or bypass the CAPTCHA.

## Verification
- full crawler suite: 12,753 passed / 113 skipped; the only initial failures were four optional provider-interface imports in the fresh venv
- after `uv sync --extra enrich`, all 4 provider-interface tests passed
- exact post-rebase focused suite: 248 passed
- Ruff check/format clean
- Pyright clean
- `git diff --check`

Relates to #7741.


viktor-shcherb:
Independent exact-SHA critic approval: `b4a6dedcb49fee835bf4536244915f6e2be1f832` — **APPROVE, no P0-P3 findings**.

The critic traced the full task-local tracker -> scrape outcome -> Redis host-circuit path, confirmed that only the typed `origin_block` protocol is promoted, a later accepted render supersedes an earlier marker, opaque challenge redirects remain sanitized and are attributed to the stable requested origin, parser/config failures remain non-transient, and cancellation behavior is unchanged. Independent checks: 244 focused tests, Ruff, formatting, Pyright, and diff check all pass.

